PR #20, opened 2018-06-14
Do not use std::to_string.

mgautierfr:
This method is not always implemented in old libstdc++.


mgautierfr:
This PR need https://github.com/kiwix/kiwix-build/pull/184, and PR https://github.com/kiwix/kiwix-build/pull/184 need this one.

Let's merge it even if it fails in the CI.

PR #30, opened 2019-03-03
Fixes around internal/external URL detection #29

### LINKED ISSUE #29: zimcheck seems to count external URL as dependencies
It should not, only things necessary to render a HTML page should be consider as a dependence, typically a node with an attribute `src=`.

kelson42:
Fixes #29

### DIFF
--- a/src/zimCheck.cpp
+++ b/src/zimCheck.cpp
@@ -326,7 +326,8 @@ int main (int argc, char **argv)
                 if (it->isRedirect() ||
                     it->isLinktarget() ||
                     it->isDeleted() ||
-                    it->getArticleSize() == 0) {
+                    it->getArticleSize() == 0 ||
+                    it->getNamespace() == 'M') {
                     continue;
                 }
                 hash_main[ adler32(it->getData()) ].push_back( it->getIndex() );
--- a/src/zimCheck.cpp
+++ b/src/zimCheck.cpp
@@ -413,6 +413,9 @@ int main (int argc, char **argv)
                 getLinks( it -> getData() , &links);
                 for(unsigned int i = 0; i < links.size(); i++)
                 {
+                    // TODO: Code in this loop is really buggy, see #32
+                    continue;
+
                     //std::cout<<"\n"<<links[i]<<std::flush;
                     links[i] = process_links( links[i] );
                     //std::cout<<"\n"<<links[i]<<std::flush;
--- a/src/zimCheck.cpp
+++ b/src/zimCheck.cpp
@@ -323,7 +323,10 @@ int main (int argc, char **argv)
             for (zim::File::const_iterator it = f.begin(); it != f.end(); ++it)
             {
                 progress.report();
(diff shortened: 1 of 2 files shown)


src/zimCheck.h
```diff
@@ -85,8 +85,9 @@ void getDependencies(std::string page, std::vector <std::string> *links)
                 {
                     i++;
                 }
-                length=i-startingPoint;
-                links->push_back(page.substr(startingPoint,length));
+                std::string url = page.substr(startingPoint, i-startingPoint);
+                if (std::regex_match(url, std::regex("http(s|)://")))
```
mgautierfr: Are you sure about this ?

`std::match` only considers full matches.
See how the following code behaves :
```cpp
#include <string>
#include <regex>
#include <iostream>
int main() {
  std::string url = "http://foo/bar.html";
  std::cout << "match " << std::regex_match(url, std::regex("http(s|)://")) << std::endl;
  std::cout << "search " << std::regex_search(url, std::regex("http(s|)://")) << std::endl;
 }
```
>>>
match 0
search 1

mgautierfr: And `std::regex` should be created only once as a "pseudo global" variable instead of being created for each url's test.

kelson42: I have renamed the PR and fixed the regex (beside other things). I have discovered an other bug which needs to be fixed too https://github.com/openzim/zim-tools/issues/32... but this is a bit more complex.

kelson42: I made it static, hope this is what you meant.

mgautierfr:
Changes in c9ac7a1d seems to overwrite changes in a5c3c316.
It seems that the only useful change in this PR is the one in c9ac7a1d where you fix the methods `isExternalUrl` and `isInternalUrl` (plus the commits where you skip empty articles or metadata articles)

src/zimCheck.cpp
```diff
@@ -409,6 +413,9 @@ int main (int argc, char **argv)
                 getLinks( it -> getData() , &links);
                 for(unsigned int i = 0; i < links.size(); i++)
                 {
+                    // TODO: Code in this loop is really buggy, see #32
+                    continue;
```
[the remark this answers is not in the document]
kelson42: @mgautierfr I'm not going to fix all bugs in the SW with this PR. This PR exists to fix the wrong external dependences detection. Here the pb is in the internal links check which was - and still - is wrong. This is different. The `continue` is there only to avoid the app to `segfault` (which is new after the fix I have made).

kelson42:
@mgautierfr Do you want me to squash everything in one commit? Might be more readable.

mgautierfr:
Not all, 9ee0abda9 and 762cde7db8 are independent and self contained. It's good.
But a5c3c316cd and c9ac7a1d80 are contradictory (the second revert a change in the first one), they should be squashed indeed.

mgautierfr:
This PR is superseded by #33.

PR #33, opened 2019-03-21
Fix check

### LINKED ISSUE #32: zimcheck broken internal link is buggy
The detection code seems to be buggy in manny manners:
* It assumes the URLs start with `../` which is a wrong assumption
* URL resolving (based on current article path) is not done properly

I had to deactivate it by working on #29

mgautierfr:
This PR supersede #30 and fix #32, #29.

Global organization of the tests has been changed. This avoid a lot of multiple decompression of the
articles. Time to run is divided by more than 2.


kelson42:
@mgautierfr I have run your branch and got this kind of feedback:
```
[ERROR] Redundant data found :
  s/12px.svg.png (idx 14) and s/12px.svg.png (idx 15
```
Here I see two problems:
* The two URLs are the same
* missing parenthesis at the end

mgautierfr:
On which zim file have you run the test ?
It seems that there the two article have different idx. So it may be an issue in the zim file directly (two articles with same url) and not in the code (while we could add a test to check for duplicated url)

mgautierfr:
I'm merging this. It is highly probable that the error is in the zim file itself, not in the code.
(parenthesis problem has been fixed)

PR #48, opened 2019-07-06
renaming

alnebrass:
Renaming with lowercase files ".cpp"


kelson42:
@alnebrass Few remarks:
* You should use "git mv" instead of deleting/recreating the files (I guess this is what you have done). So we won't loose the whole revision log of the file in git.
* Make a commit with a message which is a bit more explicit like "rename all *cpp files to lowercase".

PR #66, opened 2019-09-02
Api change

mgautierfr:


mgautierfr:
Need https://github.com/openzim/libzim/pull/277

src/zimrecreate.cpp
```diff
@@ -95,7 +95,7 @@ class Article : public zim::writer::Article         //Article class that will be
 
     bool shouldIndex() const
     {
-        return getMimeType().find("text/html") == 0;
+        return !isRedirect() && getMimeType().find("text/html") == 0;
```
kelson42: I thought we have decided it was important to index redirects too!

mgautierfr: We redirect the title all the time, whatever is `shouldIndex()`. `shouldIndex` is about full text indexing.

Saying that, I realize that we also index binary content (as image, and such) and we should not. But this is a issue to open on libzim side (I will do).

mgautierfr: https://github.com/openzim/libzim/issues/278

mgautierfr: I've removed this change.

PR #67, opened 2019-11-08
Add Releases to README

lidel:
This PR adds section about releases with a link to prebuilt binaries.

Makes it easier for people to try zim-tools without building them manually.


kelson42:
@lidel This is indeed an important information missing, even if this is in the github repo descripton. But might that be that creating a badge (beside the others) at the top of the `README.md` would be more user friendly?

lidel:
I think we probably want both?

Added a badge with the latest release (based on github tag), so user knows what to download when they click on it: [GitHub tag (latest SemVer)](https://download.openzim.org/release/zim-tools/)

PR #72, opened 2020-03-04
WIP Process files in order

MiguelRocha:
This PR improves the speed of zimdump tool by processing the clusters in order optimizing the cluster cache utilization and avoiding unnecessary decompress operations.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -11,7 +11,7 @@ if target_machine.system() != 'windows'
 endif
 
 executable('zimdump', 'zimdump.cpp',
-  dependencies: libzim_dep,
+  dependencies: [libzim_dep, dependency('threads')],
   install: true)
 
 executable('zimdiff', 'zimdiff.cpp',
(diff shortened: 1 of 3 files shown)


mgautierfr:
There is nothing to say on the code.

However, sorting the article per cluster would be better done in libzim (https://github.com/openzim/libzim/issues/300).

kelson42:
@MiguelRocha What should we do with this PR?

MiguelRocha:
@kelson42 @mgautierfr I've been doing some performance tests with this and I found that this PR does not introduce a significant performance advantage on my machine. 

We use a caching mechanism that is blocking a lot of threads and therefore it is very hard to take advantage of the multiple CPUs.

I have a couple of them in mind but I don't want to spend a lot of time on it so we maybe we can close this PR. What do you think?

kelson42:
We need to change a few things in the libzim itself (in particular the cache system) to be able to have at the multhreading bringing a lot. Closing this PR for the moment.

PR #79, opened 2020-04-01
Write the filename exacly like the url of the article.

MiguelRocha:
This PR will use the URL of each Article as the filename in the filesystem.

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -263,18 +263,18 @@ void ZimDumper::dumpFiles(const std::string& directory)
 #else
       ::mkdir(d.c_str(), 0777);
 #endif
-    std::string t = it->getUrl();
+    std::string url = it->getUrl();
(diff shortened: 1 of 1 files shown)


kelson42:
@MiguelRocha Is that PR ready? Please assign if "yes"... but I suspect is not because of the CI failing... but then we should probably put WIP in the title.

src/zimdump.cpp
```diff
@@ -263,7 +263,7 @@ void ZimDumper::dumpFiles(const std::string& directory)
 #else
       ::mkdir(d.c_str(), 0777);
 #endif
-    std::string t = it->getTitle();
+    std::string t = it->getUrl();
```
mgautierfr, line 266: While you are here, we may change the variable name also.
`t` is probably for `title` and it is no more a title.
(And it never was, it was the filename)

PR #80, opened 2020-04-03
Avoid dumping empty files for redirect articles

MiguelRocha:
For redirect Articles inserts meta html tag that redirects the client to the correct Article.

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -26,9 +26,52 @@
 #include <stdexcept>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <iomanip>
+#include <unistd.h>
+#include <array>
 
 #include "arg.h"
 #include "version.h"
+
+static bool isReservedUrlChar(const char c)
+{
+    constexpr std::array<char, 10> reserved = {';', ',', '/', '?', ':',
+                                               '@', '&', '=', '+', '$' };
+
+    return std::any_of(reserved.begin(), reserved.end(),
+                       [&c] (const char &elem) { return elem == c; } );
+}
+
+static bool needsEscape(const char c, const bool encodeReserved)
+{
+  if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9'))
+    return false;
+
+  if (isReservedUrlChar(c))
+    return encodeReserved;
+
+  constexpr std::array<char, 9> noNeedEscape = {'-', '_', '.', '!', '~',
+                                                '*', '\'', '(', ')' };
+
+  return not std::any_of(noNeedEscape.begin(), noNeedEscape.end(),
+                         [&c] (const char &elem) { return elem == c; } );
+}
+
+std::string urlEncode(const std::string& value, bool encodeReserved)
+{
+  std::ostringstream os;
+  os << std::hex << std::uppercase;
+  for (std::string::const_iterator it = value.begin();
+       it != value.end();
+       ++it) {
+    if (!needsEscape(*it, encodeReserved)) {
+      os << *it;
+    } else {
+      os << '%' << std::setw(2) << static_cast<unsigned int>(static_cast<unsigned char>(*it));
+    }
+  }
+  return os.str();
+}
 
 class ZimDumper
 {
@@ -400,4 +443,3 @@ int main(int argc, char* argv[])
   }
   return 0;
 }
-
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -301,11 +301,14 @@ void ZimDumper::dumpFiles(const std::string& directory)
   {
     std::string d = directory + '/' + it->getNamespace();
     if (ns.find(it->getNamespace()) == ns.end())
+    {
 #if defined(_WIN32)
       ::mkdir(d.c_str());
 #else
       ::mkdir(d.c_str(), 0777);
 #endif
+        ns.insert(it->getNamespace());
+    }
     std::string url = it->getUrl();
     std::string::size_type p;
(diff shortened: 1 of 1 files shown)


kelson42:
@MiguelRocha We need to be a bit (more?) carefull here about URL escaping IMO.

mgautierfr:
> For redirect Articles inserts meta html tag that redirects the client to the correct Article.

No, not all zim archive are about html file. On may use zim archive to store its file on its file system (as zip or tar do).
And in fact, even zim used for kiwix use redirection on files that are not html file (favicon for example)
It would be better to use symbolic link (if they are available in all os/fs)

kelson42:
@MiguelRocha @mgautierfr Few remarks on that:
* It is true that we can not assume that all articles are HTML, but we the mime-type we can know if this is the case or not.
* symlinks won't work properly for HTML because they are "transparent" for the HTML renderer and will brake the rendering if the redirect points to another article with a different directory depth
* I propose to use HTML redirect (like done so far) if HTML content and symlinks otherwise
* There is proper urlencode piece of code somewhere else (I believe this is in libkiwix). Not happy to see an other one appearing here.

mgautierfr:
> symlinks won't work properly for HTML because they are "transparent" for the HTML renderer and will brake the rendering if the redirect points to another article with a different directory depth

Yes, but in the same time, it is up to the client to handle this. This is why we have kiwix-serve. zim (and zim-tools) is content agnostic, it is to the user to be aware of the content is want to handle it and handle it correctly.

> There is proper urlencode piece of code somewhere else (I believe this is in libkiwix). Not happy to see an other one appearing here.

zim-tools cannot depend of kiwix-lib

kelson42:
> > symlinks won't work properly for HTML because they are "transparent" for the HTML renderer and will brake the rendering if the redirect points to another article with a different directory depth
> 
> Yes, but in the same time, it is up to the client to handle this. This is why we have kiwix-serve. zim (and zim-tools) is content agnostic, it is to the user to be aware of the content is want to handle it and handle it correctly.

We can argue about that but I would like to have a dump which is properly exploitable without complex handling on the client side.

@lidel Maybe you could share with use your opinion from a power user perspective?

> > There is proper urlencode piece of code somewhere else (I believe this is in libkiwix). Not happy to see an other one appearing here.
> 
> zim-tools cannot depend of kiwix-lib

Yes, but having another custom implementation is the worse solution IMO.

mgautierfr:
> We can argue about that but I would like to have a dump which is properly exploitable without complex handling on the client side.

From the openzim point of view, this is properly exploitable. The can access the content of the files/articles, and you know it is a redirect.
From the kiwix point of view that use zim file to store html page, it is not. But that is the a kiwix problem. zim-tools is about zim, not kiwix.

> Yes, but having another custom implementation is the worse solution IMO.

No, you don't want zim-tools to be dependent of the whole kiwix-lib (library/server/search/downloader/...) just to use one function that do a simple replace in a string.

kelson42:
> > Yes, but having another custom implementation is the worse solution IMO.
> 
> No, you don't want zim-tools to be dependent of the whole kiwix-lib (library/server/search/downloader/...) just to use one function that do a simple replace in a string.

I have written "yes", so I agree with you. But we don't have a binary choice here between another implementation and linking to the libkiwix. We have other approaches possible like: using a library which does the job, sharing a copy of the same implementation, ...

mgautierfr:
Just copy the function should be enough. I was thinking that you prefer to link with kiwix-lib than have the implementation in zim-tools.

MiguelRocha:
@mgautierfr @kelson42  Thanks for the comments. To fix the Redirect problem i propose the following:

- Have a **generic** implementation that only knows about the zimfile and its properties and dumps content to the filesystem in a generic way.

- Have a **specialization** of the implementation above that indicates what to do with redirect articles based on the mime-type. Example:
       When { mime type == html ->  write html file with meta tag}
                 { mime type == symlink -> write symbolic link (if support available)}
                 { .... -> ... } 
We can extend this to handle other use cases in the future.

mgautierfr:
There is no type "symlink". Either the article is a redirection (and the mimetype is "redirection") or a "real" article (and the mimetype depends of the content in the article).

I propose to have zimdump create symbolic link (or equivalent) by default for redirection. And have an option to change this behavior and write html redirect in case the targeted article is a html file (and still create symbolic link for other mimetype).

lidel:
Not sure if this data point is useful, but [ipfs/distributed-wikipedia-mirror/](https://github.com/ipfs/distributed-wikipedia-mirror/) uses unpacked ZIM and assumes HTML files link to each other using relative URIs. 

As long relative links work, it should be ok, so if there is a risk of having the same HTML page at different directory depth, we will need the mode which produces HTML with fixed links instead of simply symlinking the file.

src/zimdump.cpp
```diff
@@ -347,18 +327,18 @@ void ZimDumper::dumpFiles(const std::string& directory)
     {
         auto redirectArticle = it->getRedirectArticle();
         std::string redirectUrl = redirectArticle.getUrl();
-        if (redirectArticle.getMimeType() == "text/html")
+        if (symlinkdump)
+        {
+            symlink(redirectUrl.c_str(), f.c_str());
+        }
+        else
```
mgautierfr, line 334: We must do a symlink even if symlinkdump is false for other content than html:
- Symlinkdump [default]
  => symlink for all content.
- HTMLRedirect:
  => HTML redirect for html content 
  => Symlink for other content.

MiguelRocha:
@mgautierfr I cleaned the commit history. Please have a look.

MiguelRocha:
@mgautierfr I refactored the comments.

mgautierfr:
We are good.

PR #85, opened 2020-04-08
Dump nested subdirectories

MiguelRocha:
Dumps the nested subdirectories for non-A namespace articles.

```
├── -
│   ├── favicon
│   ├── j
│   │   └── js_modules
│   │       ├── article_list_home.js
│   │       ├── ext.cite.ux-enhancements.js
│   │       ├── ext.kartographer.link.js
│   │       ├── ext.kartographer.staticframe.js

```

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -341,9 +341,9 @@ void write_to_error_directoy(const std::string& base, const std::string relpath,
 inline void write_to_file(const std::string &base, const std::string& path, const char* data, ssize_t size) {
     std::string fullpath = base + path;
 #ifdef _WIN32
-    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), NULL, 0);
+    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), NULL, 0);
     std::wstring wpath(needed_size, 0);
-    MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), &wpath[0], needed_size);
+    MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), &wpath[0], needed_size);
     auto fd = _wopen(wpath.c_str(), _O_WRONLY | _O_CREAT | _O_TRUNC, S_IWRITE);
 #else
     auto fd = open(fullpath.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -334,9 +334,7 @@ void write_to_error_directory(const std::string& base, const std::string relpath
 
 #ifdef _WIN32
     auto fullpath = std::string(base + ERRORSDIR + url);
-    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), NULL, 0);
-    std::wstring wpath(needed_size, 0);
-    MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), &wpath[0], needed_size);
+    std::wstring wpath = converter.from_bytes(fullpath);
     auto fd = _wopen(wpath.c_str(), _O_WRONLY | _O_CREAT | _O_TRUNC, S_IWRITE);
 
     if (fd == -1) {
@@ -362,9 +360,7 @@ void write_to_error_directory(const std::string& base, const std::string relpath
 inline void write_to_file(const std::string &base, const std::string& path, const char* data, ssize_t size) {
     std::string fullpath = base + path;
 #ifdef _WIN32
-    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), NULL, 0);
-    std::wstring wpath(needed_size, 0);
-    MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), &wpath[0], needed_size);
+    std::wstring wpath = converter.from_bytes(fullpath);
     auto fd = _wopen(wpath.c_str(), _O_WRONLY | _O_CREAT | _O_TRUNC, S_IWRITE);
 #else
     auto fd = open(fullpath.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -24,10 +24,11 @@
 #include <zim/file.h>
 #include <zim/fileiterator.h>
 #include <stdexcept>
-#include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 #include <iomanip>
 #include <array>
+#include <vector>
 
 #include "arg.h"
 #include "version.h"
@@ -39,9 +40,31 @@
 # include <windows.h>
 #else
 # define SEPARATOR "/"
-# include <unistd.h>
-# include <sys/stat.h>
+#include <unistd.h>
 #endif
+
+#define ERRORSDIR "_errordir/"
+
+inline static void createdir(const std::string &path, const std::string &base)
+{
+    if (path.size() <= 1)
+        return ;
+
+    std::size_t position = 0;
+    while(position != std::string::npos)
+    {
+        position = path.find('/', position+1);
+        if (position != std::string::npos)
+        {
+            std::string fulldir = base + path.substr(0, position);
+            #if defined(_WIN32)
+              ::mkdir(fulldir.c_str());
+            #else
+              ::mkdir(fulldir.c_str(), 0777);
+            #endif
+        }
+    }
+}
 
 static bool isReservedUrlChar(const char c)
 {
@@ -296,26 +319,50 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
   }
   std::cout << std::endl;
 }
+void write_to_error_directoy(const std::string& base, const std::string relpath, const char *content, size_t size)
+{
+    createdir(ERRORSDIR, base);
+    std::string url = relpath;
 
-inline void write_to_file(const std::string& path, const char* data, ssize_t size) {
+    std::string::size_type p;
+    while ((p = url.find('/')) != std::string::npos)
+        url.replace(p, 1, "%2f");
+
+    std::ofstream stream(base + ERRORSDIR + url);
+
+    stream.write(content, size);
+
+    if (stream.fail() || stream.bad())
+    {
+        std::cerr << "Error writing file to errors dir. " << (base + ERRORSDIR + url) << std::endl;
+    }
+}
+
+inline void write_to_file(const std::string &base, const std::string& path, const char* data, ssize_t size) {
+    std::string fullpath = base + path;
 #ifdef _WIN32
     auto needed_size = MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), NULL, 0);
     std::wstring wpath(needed_size, 0);
     MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), &wpath[0], needed_size);
     auto fd = _wopen(wpath.c_str(), _O_WRONLY | _O_CREAT | _O_TRUNC, S_IWRITE);
 #else
-    auto fd = open(path.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
+    auto fd = open(fullpath.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
                               S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
 #endif
     if (fd == -1) {
-      throw std::runtime_error("Error opening file " + path);
+        std::cerr << "Error opening file " + fullpath + " cause: " + ::strerror(errno) << std::endl;
+        close(fd);
+        write_to_error_directoy(base, path, data, size);
+        return ;
     }
     if (write(fd, data, size) != size) {
       close(fd);
-      throw std::runtime_error("Error writing to file " + path);
+      std::cerr << "Failed writing: " << fullpath << " - " << ::strerror(errno) << std::endl;
+      write_to_error_directoy(base, path, data, size);
     }
     close(fd);
 }
+
 
 void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
 {
@@ -326,31 +373,50 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
   ::mkdir(directory.c_str(), 0777);
 #endif
 
-  std::set<char> ns;
+  std::vector<std::string> pathcache;
+  std::set<char> nscache;
   for (zim::File::const_iterator it = pos; it != file.end(); ++it)
   {
-    std::string d = directory + SEPARATOR + it->getNamespace();
-    if (ns.find(it->getNamespace()) == ns.end())
+    char filenamespace = it->getNamespace();
+    std::string d = directory + SEPARATOR + filenamespace;
+    std::string base = d + SEPARATOR;
+    if (nscache.find(filenamespace) == nscache.end())
     {
-#if defined(_WIN32)
+    #if defined(_WIN32)
       ::mkdir(d.c_str());
-#else
-      ::mkdir(d.c_str(), 0777);
-#endif
-        ns.insert(it->getNamespace());
+    #else
+      ::mkdir(base.c_str(), 0777);
+    #endif
+        nscache.insert(it->getNamespace());
     }
+    std::string t = it->getTitle();
     std::string url = it->getUrl();
-    std::string::size_type p;
-    while ((p = url.find('/')) != std::string::npos)
-      url.replace(p, 1, "%2f");
-    if ( url.length() > 255 )
+
+    auto position = url.find_last_of('/');
+    if (position != std::string::npos)
     {
-      std::ostringstream sspostfix, sst;
-      sspostfix << (++truncatedFiles);
-      sst << url.substr(0, 254-sspostfix.tellp()) << "~" << sspostfix.str();
-      url = sst.str();
+        std::string path = url.substr(0, position);
+        if (find(pathcache.begin(), pathcache.end(), path) == pathcache.end())
+        {
+            createdir(url, base);
+            pathcache.push_back(path);
+        }
     }
-    std::string f = d + SEPARATOR + url;
+
+    if ( t.length() > 255 )
+    {
+        std::ostringstream sspostfix, sst;
+        sspostfix << (++truncatedFiles);
+        sst << url.substr(0, 254-sspostfix.tellp()) << "~" << sspostfix.str();
+        url = sst.str();
+    }
+
+    std::stringstream ss;
+    ss << filenamespace;
+    ss << SEPARATOR + url;
+    std::string relative_path = ss.str();
+    std::string full_path = directory + SEPARATOR + relative_path;
+
     if (it->isRedirect())
     {
         auto redirectArticle = it->getRedirectArticle();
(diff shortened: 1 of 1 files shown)


MiguelRocha:
@mgautierfr can you please review?

src/zimdump.cpp
```diff
@@ -253,21 +271,47 @@ void ZimDumper::dumpFiles(const std::string& directory)
...
+        {
+            std::string path = t.substr(0, position);
+            if (find(pathcache.begin(), pathcache.end(), path) == pathcache.end())
+            {
+                std::string base = directory + '/' + filenamespace + '/';
+                createdir(t, base);
+                pathcache.push_back(path);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
MiguelRocha: @mgautierfr the issue here is that the filenames contain valid slashes "/". So it is impossible to know if the '/' is a directory or if its just part of the name.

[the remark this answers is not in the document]
kelson42: @MiguelRocha I see no reason why the article with URL `AC/DC` shoud not be written in the path `AC/DC`: `AC` being a directory and `DC` a file with the HTML. By doing the `/` encoding you are breaking the HTML rendering, see http://library.kiwix.org/wikipedia_en_all_nopic/A/AC/DC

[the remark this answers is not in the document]
mgautierfr: A "/" is used as a separator when serving the content through a web server.
The links are relative and take into account "/" in the path (the target of the link and the "position" of the source article").

This is why we do html redirection instead of symlink when we dump articles.
This is to ensure that the relative link is correct base on the source article.

We must follow the directory structure exposed as article path to ensure that a static file server correctly find articles.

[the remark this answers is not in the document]
MiguelRocha: > @MiguelRocha I see no reason why the article with URL `AC/DC` shoud not be written in the path `AC/DC`: `AC` being a directory and `DC` a file with the HTML. By doing the `/` encoding you are breaking the HTML rendering, see http://library.kiwix.org/wikipedia_en_all_nopic/A/AC/DC

This was causing me some confusion but now its clear.

@mgautierfr thanks for the explanation. I will do this change.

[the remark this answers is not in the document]
MiguelRocha: I'm having an issue now where the redirect articles have the same name as some sub-directories. That makes it impossible to create the sub-directories in 'A' namespace. There may be an issue with the way we write redirect articles.

mgautierfr:
I'm not sure a this PR status, but we are not c++17.

MiguelRocha:
@mgautierfr @kelson42 Now zimpdump dumps all the nested directories. When an error occurs it will log the errors to stderr and try to write the files in another directory (_errorsdir).

MiguelRocha:
@mgautierfr can you please review?

src/zimdump.cpp
```diff
@@ -296,27 +319,51 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
...
+    {
+        std::cerr << "Error writing file to errors dir. " << (base + ERRORSDIR + url) << std::endl;
+    }
+}
 
-inline void write_to_file(const std::string& path, const char* data, ssize_t size) {
+inline void write_to_file(const std::string &base, const std::string& path, const char* data, ssize_t size) {
+    std::string fullpath = base + path;
 #ifdef _WIN32
     auto needed_size = MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), NULL, 0);
```
mgautierfr, line 344: Shouldn't we use `fullpath` also on windows ?

src/zimdump.cpp
```diff
@@ -39,10 +40,32 @@
...
+
+    std::size_t position = 0;
+    while(position != std::string::npos)
+    {
+        position = path.find('/', position+1);
+        if (position != std::string::npos)
+        {
+            std::string fulldir = base + path.substr(0, position);
+            #if defined(_WIN32)
+              ::mkdir(fulldir.c_str());
```
mgautierfr, line 61: We use wide variants on windows to handle non ascii char in path.

src/zimdump.cpp
```diff
@@ -296,27 +319,51 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
...
+void write_to_error_directoy(const std::string& base, const std::string relpath, const char *content, size_t size)
+{
+    createdir(ERRORSDIR, base);
+    std::string url = relpath;
+
+    std::string::size_type p;
+    while ((p = url.find('/')) != std::string::npos)
+        url.replace(p, 1, "%2f");
+
+    std::ofstream stream(base + ERRORSDIR + url);
```
mgautierfr, line 331: `std::ofstream` doesn't handle non ascii path.
This is why we use the C `_wopen` in `write_to_file`. We must do the same in `write_to_error_directory`. (And even better, reuse the function `write_to_file` with the correct "error path" as argument).

mgautierfr, line 331: The `base` string contains the namespace (`<currentdir>/A/`).
It means that the error directory will be something like `<currentdir>/A/_errordir/`. It should be `<currentdir>/_errordir/A`

src/zimdump.cpp
MiguelRocha: Yes

src/zimdump.cpp
MiguelRocha: @mgautierfr we can use std::wofstream. It handles widechars.

src/zimdump.cpp
MiguelRocha: @mgautierfr I'm not sure on the necessity of using widechars in this case. The directory is taken from the arguments as a 8bit char array. We can have non ascii chars there.

src/zimdump.cpp
mgautierfr: The problem is not the content of the file (we can use standard char) but the path of the file. And there is no constructor accepting a `wchar`/`wstring` for the filename.

MiguelRocha: @mgautierfr `base` string does not contain the namespace. It comes directly from the program argument.

MiguelRocha: I see. But in this case, when we are creating a file we are not rendering text right? I think we should not care much about the encoding (UTF-8 on linux or UTF-16 on windows). Isn't windows taking a `char *` and do the internal encoding conversion?

src/zimdump.cpp
mgautierfr: We already have a lot of issue with this on windows. It is far better to use widechar.

MiguelRocha: Ok. I'll use widechars now and we can assess the issues on windows later.

If you agree I prefer to use a standard way of doing the conversion using stl's:
[wstring_convert](https://en.cppreference.com/w/cpp/locale/wstring_convert)

src/zimdump.cpp
mgautierfr: >  when we are creating a file we are not rendering text right?

I don't understand "rendering text".

>  Isn't windows taking a char * and do the internal encoding conversion?

No. And even if it was the case, the url is utf8, to we must correctly convert it.
The main idea in all libzim/kiwix project is to use utf8 everywhere. On windows, we must convert what we get from the system to utf8 just after "reading" it. On convert to widechars just before passing it to windows.

mgautierfr: Isn't set with the namespace here : https://github.com/openzim/zim-tools/blob/create_dir_structure_when_dumping/src/zimdump.cpp#L381-L382 ?

mgautierfr: My bad, you are using `directory + SEPARATOR` as argument. We are good here.

src/zimdump.cpp
```diff
@@ -296,27 +319,51 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
   }
   std::cout << std::endl;
 }
+void write_to_error_directoy(const std::string& base, const std::string relpath, const char *content, size_t size)
```
mgautierfr, line 322: It's missing a `r` to `directory`

src/zimdump.cpp
MiguelRocha: By rendering I mean printing text on a screen. By doing that we can have more issues regarding locale. But yes, your right. If we used C++17 or C++20 we could avoid this issues :)

 [u8path](https://en.cppreference.com/w/cpp/filesystem/path/u8path)

src/zimdump.cpp
mgautierfr: It seems that `wstring_convert` is deprecated in C++17.

src/zimdump.cpp
mgautierfr: The problem is not about rendering, but interaction with outside (printing it or passing it to the system).

> If we used C++17 or C++20 we could avoid this issues :)

Yes, but we don't :)

src/zimdump.cpp
MiguelRocha: yes because in C++17 we can use other stuff

kelson42:
@MiguelRocha Could we finally move on this PR? Would be really great to achieve to merge it before tomorrow.

MiguelRocha:
@mgautierfr  I implemented the changes that you suggested for the windows wide char paths

MiguelRocha:
@kelson42 I fixed the issue with the redirects by adding a JS redirect instruction. Everything seems to be loading fine. I used firefox launched in the output directory to test.

src/zimdump.cpp
```diff
@@ -444,9 +444,12 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
         std::string redirectUrl = redirectArticle.getUrl();
         if (symlinkdump == false && redirectArticle.getMimeType() == "text/html")
         {
+            auto encodedurl = urlEncode(redirectUrl, true);
             std::ostringstream ss;
             ss << ("<meta http-equiv=\"refresh\" content=\"0\"; url=\"" +
-                    urlEncode(redirectUrl, true) + "\" />");
+                    encodedurl + "\" />");
+            ss << ("<script type=\"text/javascript\"> window.location.href = \"" +
```
kelson42, line 451: @MiguelRocha Why we need to do so? Why a standard HTML redirect does not work?

MiguelRocha: @kelson42 firefox is not behaving correctly with the meta redirect tag. I'll test with other browsers but my bet is that not all browsers implement this specific tag

kelson42: @MiguelRocha I think I had tested with Chromium as well and it was not better. I hardly believe we face a bug in Firefox here.

kelson42:
@MiguelRocha Another problem which is unrelated I believe, is that the HTML of the redirect is incorrect. It should be something like this:
```
<!DOCTYPE html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
<meta http-equiv="refresh" content="0"; url="Code_de_d%C3%A9ontologie_m%C3%A9dicale" />
<head>
<body></body>
</html>
```

src/zimdump.cpp
kelson42: @MiguelRocha I had a look and the problem is just that your HTML is not valid. It should be here:
```
<!DOCTYPE html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
<meta http-equiv="refresh" content="0;url=Code_de_d%C3%A9ontologie_m%C3%A9dicale" />
<head>
<body></body>
</html>
```

src/zimdump.cpp
```diff
@@ -296,91 +323,149 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
...
+
+    std::string::size_type p;
+    while ((p = url.find('/')) != std::string::npos)
+        url.replace(p, 1, "%2f");
 
-inline void write_to_file(const std::string& path, const char* data, ssize_t size) {
 #ifdef _WIN32
-    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, path.data(), path.size(), NULL, 0);
+    auto fullpath = std::string(base + ERRORSDIR + url);
+    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), NULL, 0);
```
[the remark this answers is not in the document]
MiguelRocha: Done

src/zimdump.cpp
```diff
@@ -39,10 +41,35 @@
...
+    if (path.size() <= 1)
+        return ;
+
+    std::size_t position = 0;
+    while(position != std::string::npos)
+    {
+        position = path.find('/', position+1);
+        if (position != std::string::npos)
+        {
+            std::string fulldir = base + path.substr(0, position);
```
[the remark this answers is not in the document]
MiguelRocha: I tend to prefer auto also but in this case since we are doing type conversion I find it more readable to not use auto and have the types.

src/zimdump.cpp
```diff
@@ -296,91 +323,149 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
...
+    std::string fullpath = base + path;
+#ifdef _WIN32
+    auto needed_size = MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), NULL, 0);
+    std::wstring wpath(needed_size, 0);
+    MultiByteToWideChar(CP_UTF8, 0, fullpath.data(), fullpath.size(), &wpath[0], needed_size);
+    auto fd = _wopen(wpath.c_str(), _O_WRONLY | _O_CREAT | _O_TRUNC, S_IWRITE);
+#else
+    auto fd = open(fullpath.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
                               S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
 #endif
```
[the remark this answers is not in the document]
MiguelRocha: Ok we can move the code to tools after the PR is merged

src/zimdump.cpp
MiguelRocha: Thanks. I did it and it works

src/zimdump.cpp
```diff
@@ -296,91 +323,149 @@ void ZimDumper::listArticleT(const zim::Article& article, bool extra)
...
     }
-    std::string f = d + SEPARATOR + url;
+
+    if ( t.length() > 255 )
+    {
+        std::ostringstream sspostfix, sst;
+        sspostfix << (++truncatedFiles);
+        sst << url.substr(0, 254-sspostfix.tellp()) << "~" << sspostfix.str();
+        url = sst.str();
+    }
```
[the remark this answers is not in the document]
MiguelRocha: Lets do it in another PR

MiguelRocha:
> ```
> $ cat out/A/index
> <!DOCTYPE html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="0;url=Utilisateur%3ADoc_James%2FOpen_Textbook_of_Medicine" /><head><body></body></html>
> ```
> 
> `/` should not be escaped in URL otherwise the whole rendering is broken.
> 
> Otherwise I get this kind of errors:
> 
> ```
> Error opening file out/A/Fenfluramine/phentermine cause: Not a directory
> Error opening file out/A/Pipéracilline/tazobactam cause: Not a directory
> Error opening file out/A/Plante_médicinale cause: Is a directory
> Error opening file out/A/Efavirenz/emtricitabine/ténofovir cause: Not a directory
> Error opening file out/A/Amoxicilline cause: Is a directory
> ```
> 
> I don't really believe this is an "Error" because this is handled. I believe a better kind of msg would be:
> 
>     * Clearer (for the user) about the problem (here the problem is just that this file can not be written
> 
>     * This is handled and normal, so not an error
> 
>     * Where the new path is should be written
> 
>     * If we adopt something like TSV sturcture (with \t) then this is parseable
> 
>     * I would rename `_errordir` in `_exceptions`
> 
> 
> ```
> Exception\tout/A/Fenfluramine/phentermine IMPOSSIBLE out/_errordir/A%2fFenfluramine%2fphentermine
> ```

@kelson I think these are different problems.

I'm not escaping the `/` in url because I'm doing the url encoding to all link as we agreed on other PR. I've tested with firefox and it does not seem to be an issue (the browser is able to redirect).

The following error:
` Error opening file out/A/Fenfluramine/phentermine cause: Not a directory`
Happens because there are 2 articles (not directories): One is A/Fenfluramine and the other is A/Fenfluramine/phentermine. The first is written in the filesystem and then the second cannot be written in that location because it is not possible to create the directory A/Fendluramine. So it is written to _errors dir.

I can make the error msg more user friendly and change the directory name. I didn't understand the TSV structure for the msg, can you please clarify?

kelson42:
> I'm not escaping the `/` in url because I'm doing the url encoding to all link as we agreed on other PR. I've tested with firefox and it does not seem to be an issue (the browser is able to redirect).

Yes. The redirect is OK, the proper page is loaded, but all the redering is broken. You should not encode `/` in any redirect URL.

kelson42:
> The following error:
> ` Error opening file out/A/Fenfluramine/phentermine cause: Not a directory`
> Happens because there are 2 articles (not directories): One is A/Fenfluramine and the other is A/Fenfluramine/phentermine. The first is written in the filesystem and then the second cannot be written in that location because it is not possible to create the directory A/Fendluramine. So it is written to _errors dir.

I understand all of this. The point is that the error should be understandable/useful for the user lambda. If you think how it is currently (the error message) is the best then keep it, but please write as well where is the data now.

> I can make the error msg more user friendly and change the directory name. I didn't understand the TSV structure for the msg, can you please clarify?

If the user wants to deal with files in `_errordir`, the problem is that beside the log, there is no real way to know what happened exactly. If the log is parseable (like a tab separated fields). Then it can be used by a script.

MiguelRocha:
> > The following error:
> > ` Error opening file out/A/Fenfluramine/phentermine cause: Not a directory`
> > Happens because there are 2 articles (not directories): One is A/Fenfluramine and the other is A/Fenfluramine/phentermine. The first is written in the filesystem and then the second cannot be written in that location because it is not possible to create the directory A/Fendluramine. So it is written to _errors dir.
> 
> I understand all of this. The point is that the error should be understandable/useful for the user lambda. If you think how it is currently (the error message) is the best then keep it, but please write as well where is the data now.

I changed the msg to:
`Wrote out/A/Lopinavir to out/_exceptions/A%2fLopinavir`

> > I can make the error msg more user friendly and change the directory name. I didn't understand the TSV structure for the msg, can you please clarify?
> 
> If the user wants to deal with files in `_errordir`, the problem is that beside the log, there is no real way to know what happened exactly. If the log is parseable (like a tab separated fields). Then it can be used by a script.

A script can now parse like this: Wrote <problematic-file> to <newpath>

I also "unescaped" the `/` for redirects.

@mgautierfr @kelson42 It seems to be ready to merge. Let me know your comments

kelson42:
@MiguelRocha This looks really good but I would exactly happen if we have a filename with a space? Would the output still be parsable? I believe not.

MiguelRocha:
> @MiguelRocha This looks really good but I would exactly happen if we have a filename with a space? Would the output still be parsable? I believe not. This is just a proposal, but maybe we can just write a tsv file in `_exceptions`? So you would be free to write wathever you want on the STDERR.

I think is always parsable (depending on the parser ofc). IMO the filename should be everything between the string "Wrote " and " to ". That way the filename can have spaces.

I can write an additional TSV file but what if the filename also has tabs? We have the same problem.

kelson42:
> > @MiguelRocha This looks really good but I would exactly happen if we have a filename with a space? Would the output still be parsable? I believe not. This is just a proposal, but maybe we can just write a tsv file in `_exceptions`? So you would be free to write wathever you want on the STDERR.
> 
> I think it is always parsable (depending on the parser ofc). IMO the filename should be everything between the string "Wrote " and the last " to " in the line. That way the filename can have spaces.
> 
> I can write an additional TSV file but what if the filename also has tabs? We have the same problem.

There is far less chances that the title has tabs than spaces... but in that case it should be escaped. This is a standard behaviour of a CSV/TSV exporter (to escape the separator character in fields).

MiguelRocha:
> > > @MiguelRocha This looks really good but I would exactly happen if we have a filename with a space? Would the output still be parsable? I believe not. This is just a proposal, but maybe we can just write a tsv file in `_exceptions`? So you would be free to write wathever you want on the STDERR.
> > 
> > 
> > I think it is always parsable (depending on the parser ofc). IMO the filename should be everything between the string "Wrote " and the last " to " in the line. That way the filename can have spaces.
> > I can write an additional TSV file but what if the filename also has tabs? We have the same problem.
> 
> There is far less chances that the title has tabs than spaces... but in that case it should be escaped. This is a standard behaviour of a CSV/TSV exporter (to escape the separator character in fields).

Ok. Should we do this on a separated PR since it seems to be an specific enhancement?

- TSV filename?
- What should be the column names (header)?

kelson42:
@MiguelRocha Please open a ticket wich explain the problem and we should merge this PR.

kelson42:
@MiguelRocha I have open a ticket to keep track here https://github.com/openzim/zim-tools/issues/108... So we only need @mgautierfr approval here right?

src/zimdump.cpp
```diff
@@ -39,13 +41,38 @@
 # include <windows.h>
 #else
 # define SEPARATOR "/"
-# include <unistd.h>
-# include <sys/stat.h>
+#include <unistd.h>
 #endif
 
+#define ERRORSDIR "_exceptions/"
```
mgautierfr, line 47: We have to agree with the doc in #89 
But I don't know what is ok (the code or the doc)

MiguelRocha:
@kelson42 thanks for opening the ticket. @mgautierfr since this PR is ready can we merge it before the usage PR?

src/zimdump.cpp
kelson42: The code is OK, I ask for this name `_exceptions`.

kelson42:
@mgautierfr @MiguelRocha I have rebase and will merged as soon as the CI has passed.

Issue #70, opened 2020-01-14
Rewrite API of zimdump.


mgautierfr:
Zimdump is a pretty old tool. It was mainly used a debug tools. Its API is very odd and performance is the last of this feature.

There is a lot of issue about zimdump. Some about the odd API (#63, #4), some about using zimdump as a "real" tool and not just a debug tool (#69, #24, #68).
In both case, we should rethink the API of the tool.

There are different actions done by zimdump :
- Print zim file info (the zim header)
- Print content of an article
- Print info about an article (the dirent)
- List the content of a zim file
- Dump all articles (or a subset) in a directory
- Verify the checksum of the zim file.

I propose a API based on subcommand.
- dump
- list
- info

`dump` subcommand would be used to dump the content of article(s). Either on stdout (for one article) or in a directory (for one or several articles).

`list` would be use to list all (or a subset ?) articles (url or title). It could be use to check than a article exists.

`info` would be use to print info on the zim file or on a specific article.

The "verify checksum action" should be in `zimcheck` and will be removed from `zimdump`.

Articles could be selected by Id, Url or Title. Url would have to be "long url" (with namespace).

Some command line examples (exact option's names are not definitive) :
`zimdump <zim_file> info` Print the header of the zimfile.
`zimdump <zim_file> info --id 50` Print dirent of article 50
`zimdump <zim_file> info --url A/foo.html` Print dirent of article 'A/foo.html' (or 'foo.html' in namespace 'A')
`zimdump <zim_file> info --title Foo` Print dirent of article named "Foo"
`zimdump <zim_file> list` List all article (url) in the zim file.
`zimdump <zim_file> list --title` List all article (title) in the zim file.
`zimdump <zim_file> dump --output Foo` Dump all articles in the directory `Foo`.
`zimdump <zim_file> dump --title Foo --output Foo.html` Dump the content of the artcle named "Foo" in Foo.html.
`zimdump <zim_file> dump --url A/foo.html --output Foo.html` Dump the content of the article `A/foo.html` in Foo.html.
`zimdump <zim_file> dump --url A/foo.html --output -` Dump the content of the article `A/foo.html` on stdout.

kelson42:
@mgautierfr @MiguelRocha To me this is how it should looks like:
``` 
usage: zimdump <command> [<args>] <file>

These are common zimdump commands used in various situations:

    ls            List content
    dump          Dump content
        --dir     Write content to the filesystem (terminal otherwise)
    info          Show meta structure details

[...]

PR #89, opened 2020-04-15
New zimdump cmdline arguments.

MiguelRocha:
This PR refactors the argument parsing mechanism of zimdump. It applies a subcommand structure with the current subcommands:

- info
- dumpall
- dump
- list
- verifychecksum

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,8 +16,7 @@ libzim_dep = dependency('libzim', version : '>=6.1.1', static:static_linkage)
 
 compiler = meson.get_compiler('cpp')
 find_library_in_compiler = meson.version().version_compare('>=0.31.0')
-rt_dep = compiler.find_library('rt', required:false)
-docopt_dep = compiler.find_library('docopt', required:false)
+docopt_dep = dependency('docopt', static:static_linkage)
 
 with_writer = target_machine.system() != 'windows'
 
--- a/meson.build
+++ b/meson.build
@@ -16,6 +16,7 @@ libzim_dep = dependency('libzim', version : '>=6.1.1', static:static_linkage)
 
 compiler = meson.get_compiler('cpp')
 find_library_in_compiler = meson.version().version_compare('>=0.31.0')
+rt_dep = dependency('rt', required:false)
 docopt_dep = dependency('docopt', static:static_linkage)
 
 with_writer = target_machine.system() != 'windows'
--- a/meson.build
+++ b/meson.build
@@ -17,6 +17,7 @@ libzim_dep = dependency('libzim', version : '>=6.1.1', static:static_linkage)
 compiler = meson.get_compiler('cpp')
 find_library_in_compiler = meson.version().version_compare('>=0.31.0')
 rt_dep = compiler.find_library('rt', required:false)
+docopt_dep = compiler.find_library('docopt', required:false)
 
 with_writer = target_machine.system() != 'windows'
 
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -31,6 +31,7 @@
 #include <array>
 #include <vector>
 #include <codecvt>
+#include <unordered_map>
 
 #include "arg.h"
 #include "version.h"
@@ -368,8 +369,11 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::f
 
   std::vector<std::string> pathcache;
   std::set<char> nscache;
+  for (zim::File::const_iterator it = m_file.begin(); it != m_file.end(); ++it)
   {
     char filenamespace = it->getNamespace();
+    std::string d = directory + SEPARATOR + filenamespace;
+    std::string base = d + SEPARATOR;
     if (nsfilter(it->getNamespace()) == false)
         continue;
 
@@ -443,9 +447,7 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::f
 
 static const char USAGE[] =
 R"(
-#ifndef _WIN32
-                   "  -s        Use symlink to dump html redirect. Else create html redirect file."
-#endif
+
 Usage:
   zimdump ls [--details] [--idx=INDEX|([--url=URL] [--ns=N])] [--] <file>
   zimdump list [--details] [--idx=INDEX|([--url=URL] [--ns=N])] [--] <file>
@@ -465,6 +467,7 @@ Selectors:
 Options:
   --details    Show details about the articles. Else, list only the url of the article(s).
   --dir=DIR    Directory where to dump the article(s) content.
+  --redirect   Use symlink to dump redirect articles. Else create html redirect file
   -h, --help   Show this help
   --version    Show zimdump version.
 
@@ -497,9 +500,7 @@ void subcmdDumpAll(ZimDumper &app, std::map<std::string, docopt::value> &args, b
 
 void subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
 {
-    bool redirect = false;
-
-    if (args["--redirect"]) redirect = true;
+    bool redirect = args["--redirect"].asBool();
 
     if (args["--dir"]) {
         std::function<bool (const char c)> filter = [](const char /*c*/){return true; };
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -56,11 +56,9 @@ inline static void createdir(const std::string &path, const std::string &base)
         return ;
 
     std::size_t position = 0;
-    while(position != std::string::npos)
-    {
+    while(position != std::string::npos) {
         position = path.find('/', position+1);
-        if (position != std::string::npos)
-        {
+        if (position != std::string::npos) {
             std::string fulldir = base + path.substr(0, position);
             #if defined(_WIN32)
             std::wstring wfulldir = converter.from_bytes(fulldir);
@@ -137,7 +135,7 @@ class ZimDumper
 
     bool hasNamespace(char ns);
 
-    int dumpFiles(const std::string& directory, bool symlinkdump, std::function<bool (const char c)> nsfilter);
+    void dumpFiles(const std::string& directory, bool symlinkdump, std::function<bool (const char c)> nsfilter);
 };
 
 zim::Article ZimDumper::getArticleByUrl(const std::string &url)
@@ -158,12 +156,12 @@ bool ZimDumper::hasNamespace(char ns)
 void ZimDumper::printInfo()
 {
   std::cout << "count-articles: " << m_file.getCountArticles() << "\n";
-  if (verbose)
-  {
+  if (verbose) {
     std::string ns = m_file.getNamespaces();
     std::cout << "namespaces: " << ns << '\n';
-    for (std::string::const_iterator it = ns.begin(); it != ns.end(); ++it)
+    for (std::string::const_iterator it = ns.begin(); it != ns.end(); ++it) {
       std::cout << "namespace " << *it << " size: " << m_file.getNamespaceCount(*it) << '\n';
+    }
   }
   std::cout << "uuid: " << m_file.getFileheader().getUuid() << "\n"
                "article count: " << m_file.getFileheader().getArticleCount() << "\n"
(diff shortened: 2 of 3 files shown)


data-man:
IMO [CLI11](https://github.com/CLIUtils/CLI11) is the best command line parser I ever tried.

kelson42:
@mgautierfr @MiguelRocha I think the `verifychecksum` action should/could be removed in favour of the one in `zimcheck`. What do you think?

src/zimdump.cpp
```diff
@@ -290,114 +298,100 @@ void ZimDumper::verifyChecksum()
...
+static const char USAGE[] =
+R"(zimdump.
+
+    Usage:
+      zimdump <zim_file> info [--id=<idvalue>|--namespaceinfo=<ns>]  [-v]
+      zimdump <zim_file> dumpall --output=<outputdir>
+      zimdump <zim_file> list [--asTable|--metadata] [--extra]
+      zimdump <zim_file> dump [--url=<urlvalue>|--title=<titlevalue>|--offset=<offsetvalue>] [--namespace=<ns>] [--asPage|--asMeta] [--extra]
+      zimdump <zim_file> verifychecksum
+
```
kelson42, line 310: To me there is an empty line too much here.

kelson42:
I'm a bit skeptical here because:
* `zimdump wikipedia_en_ray_charles_maxi_2020-03.zim list --namespace=X` seems to not work (and I don't understand why?)
* It seems there is no option at all to find an article
* Most of the options are not clear to me (unexplained).
* `--asTable` does not work as it should (this is a CSV, not a table)

Probably (part of) the problem comes from an incomplete/weak requirements :(

meson.build
```diff
@@ -16,5 +16,6 @@ libzim_dep = dependency('libzim', version : '>=6.1.0', static:static_linkage)
 
 compiler = meson.get_compiler('cpp')
 rt_dep = compiler.find_library('rt', required:false)
+docopt_dep = compiler.find_library('docopt', required:false)
```
mgautierfr, line 19: docopt is required.

src/zimdump.cpp
```diff
@@ -290,114 +298,99 @@ void ZimDumper::verifyChecksum()
...
-int main(int argc, char* argv[])
+static const char USAGE[] =
+R"(zimdump.
+
+    Usage:
+      zimdump <zim_file> info [--namespaceinfo=<ns>]  [-v]
+      zimdump <zim_file> dumpall --output=<outputdir>
+      zimdump <zim_file> list [--asTable|--metadata] [--extra]
+      zimdump <zim_file> dump [--url=<urlvalue>|--title=<titlevalue>|--offset=<offsetvalue>] [--namespace=<ns>] [--asPage|--asMeta] [--extra]
+      zimdump <zim_file> verifychecksum
```
mgautierfr, line 309: We shouldn't have a verifychecksum sub-command.
zimcheck is better for that.

mgautierfr:
> zimdump wikipedia_en_ray_charles_maxi_2020-03.zim list --namespace=X seems to not work (and I don't understand why?)

There is not option `--namespace` for sub-command `list` so you cannot use it.

> --asTable does not work as it should (this is a CSV, not a table)

It is how "previous" zimdump was working.

MiguelRocha:
@mgautierfr do you agree with the implementation? Can you please review the code?

kelson42:
@MiguelRocha The whole CI is failing, the discussions are not resumed. Might you please clean the situation?

src/zimdump.cpp
```diff
@@ -393,124 +402,128 @@ void ZimDumper::verifyChecksum()
...
+      zimdump <zim_file> list [--details] [--idx=INDEX|(--url=URL [--ns=N])]
+      zimdump <zim_file> dump --dir=DIR [--ns=N] [--redirect]
+      zimdump <zim_file> dump (--idx=INDEX |(--url=URL [--ns=N]))
+
+    Options:
+      --idx=INDEX       The index of the article to list/dump
+      --ns=N            Namespace selection [default: A]
+      -h --help         Show this screen.
+      --version         Show version.
+)";
```
mgautierfr, line 419: Please use the usage text I've give in https://github.com/openzim/zim-tools/issues/70#issuecomment-636773620

src/zimdump.cpp
```diff
@@ -393,124 +402,128 @@ void ZimDumper::verifyChecksum()
...
+        app.printInfo();
+    }
+}
+
+void subcmdDumpAll(ZimDumper &app, std::map<std::string, docopt::value> &args, bool redirect, std::function<bool (const char c)> nsfilter)
+{
+#ifdef _WIN32
+    app.dumpFiles(args["--output"].asString(), false, nsfilter);
+#else
+    app.dumpFiles(args["--dir"].asString(), redirect, nsfilter);
```
mgautierfr, line 436: Why we use `--output` on WIN32 and `--dir` on linux/mac ?

src/zimdump.cpp
```diff
@@ -90,7 +92,13 @@ class ZimDumper
...
 
   public:
-    ZimDumper(const char* fname, bool titleSort)
+    ZimDumper(const std::string& fname)
+      : file(fname),
+        pos(file.begin()),
+        verbose(false)
+      { }
+
+    ZimDumper(const std::string fname, bool titleSort)
```
mgautierfr, line 101: You create a second creator but you never use it.
And I think we don't need it anyway.

src/zimdump.cpp
MiguelRocha: My type. Thanks

MiguelRocha:
> I think that you tried to keep the same structure and logic that the previous code but we should avoid that.
> 
> For example, you keep the "global" `pos` iterator but we shouldn't. Each method/command should do its own way. The `dump` method should use a iterator (cluster order), the `list` method should use a iterator (url order) and the list/dump with a selector should probably don't use a iterator but directly get an `Article`

Ok. I was trying to avoid making a big refactor in the code. But now, as you pointed, a lot of things start to not make sense (like the pos iterator).

MiguelRocha:
> @MiguelRocha The whole CI is failing, the discussions are not resumed. Might you please clean the situation?

The CI is now green. I'll apply the changes suggested by @mgautierfr. Meanwhile would be nice to have some input on the commands/API

mgautierfr:
> Meanwhile would be nice to have some input on the commands/API

As said in the review, please use the usage text defined in #70

src/zimdump.cpp
MiguelRocha: I removed it.

src/zimdump.cpp
```diff
@@ -87,68 +87,72 @@ std::string urlEncode(const std::string& value, bool encodeReserved)
...
 
+zim::Article ZimDumper::getArticleByUrl(const std::string &url)
+{
+    return m_file.getArticleByUrl(url);
+}
+
+zim::Article ZimDumper::getArticleByTitle(zim::size_type idx)
+{
+    return m_file.getArticleByTitle(idx);
+}
```
mgautierfr, line 122: We should use `getArticle`. (url order)

src/zimdump.cpp
```diff
@@ -385,132 +367,143 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
...
+                         true,
+                         "zimdump 1.0");
+
+    try {
+        ZimDumper app(args["<file>"].asString());
+
+        std::unordered_map<std::string, std::function<void(ZimDumper&, decltype(args)&)>> dispatchtable = {
+            {"info",            subcmdInfo },
+            {"dump",            subcmdDump },
+            {"list",            subcmdList }
```
mgautierfr, line 495: Is it missing the "ls" subcommand ? (or docopt detect it is a alias ?)

src/zimdump.cpp
```diff
@@ -327,9 +306,12 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
 #endif
 
   std::set<char> ns;
-  for (zim::File::const_iterator it = pos; it != file.end(); ++it)
+  for (zim::File::const_iterator it = m_file.begin(); it != m_file.end(); ++it)
   {
     std::string d = directory + SEPARATOR + it->getNamespace();
+    if (nsfilter(it->getNamespace()))
```
mgautierfr, line 312: It is "funny", I would have written the other way (the filter must return `true` to keep the article). As for in python filter (https://docs.python.org/3/library/functions.html#filter) and c++ copy_if (https://en.cppreference.com/w/cpp/algorithm/copy)

But it works. No problem.

src/zimdump.cpp
MiguelRocha: Done

src/zimdump.cpp
MiguelRocha: Docopt doesnt seem to detect. I added it

src/zimdump.cpp
MiguelRocha: Changed. I also think it makes a little bit more sense

MiguelRocha:
@kelson42 @mgautierfr I've implemented the changes from the code review and rebased the code. Let me know your comments.

MiguelRocha:
>     * I don't understand the difference between `list` and `ls`. Only one othem should here (`list`?).
> 
>     * I miss an introduction/description sentence(s) at the start of the usage(). Something like "Inspect and dump ZIM file content".
> 
>     * `zimdump 1.0` is not the proper version, should be the same version like other tools.
> 
>     * What does `[--]` means in `zimdump info [--] [--ns=N] <file>` (and in other usage() lines)?
> 
>     * `$ zimdump info --ns=42 vikidia_ca_all_maxi_2020-06.zim` does not return any error.
> 
>     * Because `zimdump dump` has two lines in the usage which work quite differently, I would propose to just move `zimdump dump (--idx=INDEX|(--url=URL [--ns=N])) [--] <file>` to `zimdump show (--idx=INDEX|(--url=URL [--ns=N])) [--] <file>`.
> 
> 
> @MiguelRocha Please resolve the conversation which should be resolved.

I removed "ls" alias as we agreed. All the other points are implemented. Let me know what you think

MiguelRocha:
> zimdump always return `0`. Please follow the usage text and return `!=0` if there is a error.
> 
> This pr is 21 commits long. Please squash the commits to have a commit per "logical change". If there is only one change that cannot be split, then use only one commit.

Added return errors.

I'll check the commit history and try to come up with a more readable commits but please keep in mind that since this is PR a major refactor to the code it is likely that we have a lot of commits.

MiguelRocha:
@mgautierfr I've squashed some commits. I removed commits that introduced just small changes/fixes to the code and I tried to keep the commits that introduce relevant (big) changes to the code base without losing a sense of "history".

src/zimdump.cpp
```diff
@@ -379,7 +375,7 @@ inline void write_to_file(const std::string &base, const std::string& path, cons
 }
 
 
-void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
+int ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::function<bool (const char c)> nsfilter)
```
mgautierfr, line 378: Do not make the method return a `int` if it always return 0.
It is useless and error prone (caller will assume you return something different than 0 if something goes wrong)

src/zimdump.cpp
```diff
@@ -463,134 +461,164 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
...
-    else if (url.isSet())
-      app.findArticleByUrl(std::string(url));
+    if (idx || url)
+    {
+        zim::Article article;
+        if (idx)
+        {
+            article = app.getArticle(args["--idx"].asLong());
+        } else if(url)
+        {
```
mgautierfr, line 577: It is a matter of style, but I prefer the brace after the if :
```
if (idx) {
   ...
} else if (url) {
   ...
} else {
   ...
}
```

As it is done for the `for`, `try`, ...

mgautierfr:
> I've squashed some commits. I removed commits that introduced just small changes/fixes to the code and I tried to keep the commits that introduce relevant (big) changes to the code base without losing a sense of "history".

That's good. I've check all commits individually but it what I wanted. Remove all iteration of the code review and keep different commits for different relevant changes.

src/zimdump.cpp
MiguelRocha: I agree. I left `subcmdDumpAll` to return an int thou because it is part of the signature.

src/zimdump.cpp
MiguelRocha: For me it is ok if everything follows the same convention. I tried to uniform for the other code also but I think we should have a separated PR to apply the code formatting rules (we can even have a clang-format file with our rules).

src/zimdump.cpp
```diff
@@ -463,134 +461,164 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump)
...
+
+int subcmdInfo(ZimDumper &app, std::map<std::string, docopt::value> &args)
 {
-  if (file.verify())
-    std::cout << "checksum ok" << std::endl;
-  else
-    std::cout << "no checksum" << std::endl;
+    if (args["--ns"]) {
+        char ns = args["--ns"].asString().at(0);
+        if (app.hasNamespace(ns))
```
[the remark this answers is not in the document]
MiguelRocha: done

MiguelRocha:
> The code is functional. But I've few comments about the style.
> 
> Please try to keep the check consistent.
> 
> Subcommands are globally two parts :
> 
>     * Getting what to display/to do stuff on.
> 
>     * The action on the "object".
> 
> 
> Sometime, you are checking if the article is good in the first part before doing the action, and sometime you pass the article to the method and the method check it.

I decided to check on both since the methods can be used by other callers and it isn't a relevant performance decrease. 

I've also applied your suggestion in terms of style but since there is a lot to be said in terms of formatting I think we should have a separated PR to come up with the rules.

In terms of log history I didn't understand if the logs are OK or if you want more changes. I think every commit is separated in logical changes (that include changes that were requested in the code review process).

mgautierfr:
> I've also applied your suggestion in terms of style but since there is a lot to be said in terms of formatting I think we should have a separated PR to come up with the rules.

Yes, now we start to be several to hack on openzim/kiwix, I think it is time to have a coding style and a defined workflow about the PR and how to do.

I will do that and we will see then.

> In terms of log history I didn't understand if the logs are OK or if you want more changes. I think every commit is separated in logical changes (that include changes that were requested in the code review process).

The log history is good.

mgautierfr:
@MiguelRocha Please rebase on `origin/master`

MiguelRocha:
@mgautierfr  @kelson42  done. We are ready to merge.

PR #94, opened 2020-05-07
Build zim-tools for windows on the CI.

mgautierfr:


kelson42:
Nice! Then just missing its Windows distribution in http://download.openzim.org/release/zim-tools/.

mgautierfr:
The publication of the build release has to be made on kiwix-build side.

kelson42:
@MiguelRocha @mgautierfr It is great to see progress on libzim-tools Windows support, but I want to keep the work necessary to do that under control. This is only a nice to have. This is not a priority. At zim level and in term of portability (binding, tools) the priority is to extend support of macOS. So please don't make any move implying a lot of additional work without creating first the ticket and securing there is an agreement on this.

PR #95, opened 2020-05-15
Use double braces to initialize std::array.

mgautierfr:
std::array will be initialize to an array of 10 elements (0) with
`{10, 0}`.
To initialize it with 2 elements (10 and 0), we use a double braces
`{{10, 0}}`.

To avoid mistake xcode genere a waring and as we are using `-Werror` it
breaks compilation.
(gcc remove this warnig because it was too anoying).
See https://bugs.llvm.org/show_bug.cgi?id=21629

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -44,8 +44,8 @@
 
 static bool isReservedUrlChar(const char c)
 {
-    constexpr std::array<char, 10> reserved = {';', ',', '/', '?', ':',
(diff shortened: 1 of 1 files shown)


mgautierfr:
I don't know why the macos ci (github actions) didn't fail on the PR and the travis Ci fails (https://travis-ci.com/github/kiwix/kiwix-build/jobs/334526303)

PR #104, opened 2020-06-17
Integrate zimwriterfs sources into zim-tools source tree

### LINKED ISSUE #92: Move zimwriterfs to zim-tools repo
This would simplify a few things.

asashnov:
Closes #92 
Sources of zimwriterfs are now incorporated into zim-tools build process.


kelson42:
@asashnov Ci is still failing unfortunately.

asashnov:
@kelson42 fixed, building of zimwriterfs is turned off for Windows

kelson42:
@asashnov Thx, I have assigned @mgautierfr as reviewer. In the meantime, if you could continue with the other tasks, that would be great.

asashnov:
I agree on tools.(cpp|h)

Instead of moving all the dependencies to the root, I thought about making zimwriterfs with its dependencies optional. So the developers who work on zim-tools don't need to install extra dependencies after we merge the change.

kelson42:
> I thought about making zimwriterfs with its dependencies optional.

zimwriterfs should not be optional.

mgautierfr:
> Instead of moving all the dependencies to the root, I thought about making zimwriterfs with its dependencies optional. So the developers who work on zim-tools don't need to install extra dependencies after we merge the change.

The idea is that zimwriterfs IS part of the tools.
As we don't compile it on windows, the dependencies should be optional on windows. But on linux, zimwriterfs has to be compiled.

One other thing to do is to adapt kiwix-build to not build the (independent) zimwriterfs project and package (subtools) zimwriterfs with the zim-tools archive.

asashnov:
Dependencies are moved to the root meson.build, but they are optional:
* when zimwriterfs is not turned off,
* and the platform is not Windows.

kelson42:
@asashnov If you are finished with the PR improvements, please:
* Close the discussion thread which have to to closed (otherwise ask for additional info/feebback)?
* Request a new review.

asashnov:
> The merge is globally good.
> However I have few remarks :
> 
>     * I prefer to have all the dependencies "detection" in the main meson.build file (not in the deeper `src/zimwriterfs/meson.build`.
> 
>     * Maybe, as the resultant `src/zimwriterfs/meson.build` will be pretty small, we can simply move the code in `src/meson.build`. But I'm not totally sure about that.
> 



> The merge is globally good.
> However I have few remarks :
> 
>     * I prefer to have all the dependencies "detection" in the main meson.build file (not in the deeper `src/zimwriterfs/meson.build`.
> 
>     * Maybe, as the resultant `src/zimwriterfs/meson.build` will be pretty small, we can simply move the code in `src/meson.build`. But I'm not totally sure about that.
> 
>     * The same way, we may move the `tools.(cpp|h)` file in `src`. Functions there (getLinks, path manipulation) could be used by other tools (zimdump, zimcheck).

The changes are implemented and rebased on top of the current master.

meson_options.txt
```diff
@@ -2,3 +2,5 @@ option('static-linkage', type : 'boolean', value : false,
   description : 'Create statically linked binaries.')
 option('magic-install-prefix', type : 'string', value : '',
   description : 'Prefix where libmagic has been installed')
+option('with-zimwriterfs', type : 'boolean', value : false,
```
mgautierfr: Please turn it on directly in this commit not in the next one.
It is useless (but "pollute" the git history) to do this in two steps.

meson.build
```diff
@@ -14,10 +14,51 @@ endif
...
 
-with_writer = get_option('with-zimwriterfs')
-
 compiler = meson.get_compiler('cpp')
 find_library_in_compiler = meson.version().version_compare('>=0.31.0')
 rt_dep = compiler.find_library('rt', required:false)
 
+with_writer = get_option('with-zimwriterfs')
+
+if with_writer and target_machine.system() != 'windows'
```
mgautierfr, line 23: Maybe you could do :
```
with_writer = get_option('with-zimwriterfs') and target_machine.system() != 'windows'

if with_writer
  ...
```

This way you can simply just test `with_writer` when you include the subdirectory zimwriterfs later.

README.md
```diff
@@ -90,8 +90,8 @@ will be created in the "build" directory created automatically by
 Meson. If you want statically linked libraries, you can add
 `-Dstatic-linkage=true` option to the Meson command.
 
-If you also want to build zimwriterfs tool (which requires some extra
-dependencies) add `-Dwith-zimwriterfs=true` option.
+If you want to disable building zimwriterfs (to not install unneeded
+dependencies if you not need it) add `-Dwith-zimwriterfs=false` option.
```
mgautierfr, line 94: Please squash this commit directly with the commit adding the option.

meson.build
asashnov: Good point, done: 'if with_writer' condition used in both cases.

meson_options.txt
asashnov: optionality of zimwriterfs removed according to https://github.com/openzim/zim-tools/pull/104#pullrequestreview-436483298

README.md
asashnov: Done, commits are revisited:

0b13217e567f  - droped ( update README - will not change, no new option added)
cba95763f08d0 -  edited (moving build dependencies to root, turning off on Windows)
4692603b6f46ebd4  - droped (making writer optional)
1ad717098898  - kept (Moving tools.(cpp|h) to src)
fed07899802ca12  - droped (turning off building on Windows)
4fe0d367eb88  - kept (Integrate sources into tree)

meson.build
```diff
@@ -15,6 +15,50 @@ endif
...
 
 compiler = meson.get_compiler('cpp')
+find_library_in_compiler = meson.version().version_compare('>=0.31.0')
 rt_dep = compiler.find_library('rt', required:false)
 
+with_writer = target_machine.system() != 'windows'
+
+if with_writer
+  thread_dep = dependency('threads')
+  libzim_dep = dependency('libzim', version : '>=6.1.2', static:static_linkage)
```
mgautierfr, line 25: libzim is already detected globally.

meson.build
```diff
@@ -15,6 +15,50 @@ endif
...
+  zlib_dep = dependency('zlib', static:static_linkage)
```
mgautierfr, line 26: zlib is used by `tools.cpp` and it is in the "global" src directory.
It is indeed compiled only if we compile zimwriterfs (everywhere but windows) but as it is in the global src, it will ends to be compiled for other binaries. As zlib is a dependency of libzim, it will be always available. We can move this test "globally".

In fact, this is the same for gumbo and magic. Maybe moving the tools.cpp in the global directory was not a good idea :/

asashnov: Actually from the entire tools.(h|cpp)  libmagic is used by `getMimeTypeForFile()`   and zlib is used by `inflateString()`
We can move those two function into src/zimwriterfs/tools.cpp, leaving the rest in src/tools.cpp

asashnov: @mgautierfr What you think about splitting tools.cpp ?

asashnov: removed the duplicate of libzim_dep; splitted tools.(h|cpp) to src/tools.(h|cpp) and src/zimwriterfs/tools.(h|cpp) with functions which depends on zlib, magic and gumbo.

kelson42:
@mgautierfr Can we merge that one?

PR #106, opened 2020-06-20
Better support of WebP image file format

kelson42:


mgautierfr:
I would have wait for #104 to be merge to avoid conflict.

PR #113, opened 2020-07-06
Update zimwriterfs Docker deps

kelson42:
@mgautierfr I really don't understand why I have to put `export PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig/`, but your are the pkg-config expert :)


rgaudin:
Hum, I doubt it was tested :)
Fixed it but we still need the `LD_LIBRARY_PATH`

PR #117, opened 2020-07-14
Fix crash in zimcheck

### LINKED ISSUE #91: Zimcheck crach scenario
$ zimcheck edutechwiki_fr_all_maxi_2020-05.zim 
[INFO] Checking zim file edutechwiki_fr_all_maxi_2020-05.zim
[INFO] Verifying Internal Checksum.. 
  [INFO] Internal checksum found correct
[INFO] Searching for metadata entries..
[INFO] Searching for Favicon..
[INFO] Searching for main page..
[INFO] Verifying Articles' content.. 
basic_string::_M_replace_aux
$ echo $?
2
```

### PR DESCRIPTION
**Pull request by MiguelRocha**
This PR fixes a crash in zimcheck while validating the links of a particular Article #91 

1. Adds _Out of Bounds_ validation for links. i.e: `../../../../file.js` and base `A`
2. Adds a slight enhancement in the way zimcheck validates links by removes duplicates from the list of links to validate

Fixes #91

### DIFF
--- a/src/zimcheck.cpp
+++ b/src/zimcheck.cpp
@@ -370,7 +370,6 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
     // So all article with the same hash will be stored in the same list.
     std::map<unsigned int, std::list<zim::article_index_type>> hash_main;
 
-    std::string previousLink;
     int previousIndex = -1;
 
     progress.reset(f.getFileheader().getArticleCount());
@@ -413,28 +412,36 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
             baseUrl.resize( pos==baseUrl.npos ? 0 : pos );
 
             auto links = getLinks(it->getData());
-            for(auto olink: links)
+            std::unordered_map<std::string, std::vector<std::string>> filtered;
+            for (const auto &l : links)
             {
-                if (olink.front() == '#' || olink.front() == '?')
-                    continue;
-                if (isInternalUrl(olink)) {
-                    auto link = normalize_link(olink, baseUrl);
-                    char nm = link[0];
-                    std::string shortUrl(link.substr(2));
-                    auto a = f.getArticle(nm, shortUrl);
-                    if (!a.good())
+                if (l.front() == '#') continue;
+                if (isInternalUrl(l) == false) continue;
+
+                auto normalized = normalize_link(l, baseUrl);
+                filtered[normalized].push_back(l);
+            }
+
+            for(const auto &p: filtered)
+            {
+                std::string link = p.first;
+                char nm = link[0];
+                std::string shortUrl(link.substr(2));
+                auto a = f.getArticle(nm, shortUrl);
+                if (!a.good())
+                {
+                    int index = it->getIndex();
+                    if (previousIndex != index)
                     {
-                        int index = it->getIndex();
-                        if ((previousLink != link) && (previousIndex != index) )
-                        {
-                            std::ostringstream ss;
-                            ss << link << " (" << olink << ") was not found in article " << it->getLongUrl();
-                            reporter.addError(TestType::URL_INTERNAL, ss.str());
-                            previousLink = link;
-                            previousIndex = index;
-                        }
-                        reporter.setTestResult(TestType::URL_INTERNAL, false);
+                        std::ostringstream ss;
+                        ss << "The following links:\n";
+                        for (const auto &olink : p.second)
+                            ss << "- " << olink << '\n';
+                        ss << "(" << link << ") were not found in article " << it->getLongUrl();
+                        reporter.addError(TestType::URL_INTERNAL, ss.str());
+                        previousIndex = index;
                     }
+                    reporter.setTestResult(TestType::URL_INTERNAL, false);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/zimcheck.cpp [hunk 1]
```diff
@@ -398,24 +418,32 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
...
                     auto a = f.getArticle(nm, shortUrl);
                     if (!a.good())
                     {
                         int index = it->getIndex();
-                        if ((previousLink != link) && (previousIndex != index) )
+                        if (previousIndex != index)
                         {
                             std::ostringstream ss;
-                            ss << link << " (" << olink << ") was not found in article " << it->getLongUrl();
+                            ss << link << " (" << olink << ") was not found in article " << it->getLongUrl() << " | shortUrl: " << shortUrl << " ns: " << nm;
```
mgautierfr, line 445: The plan is to remove the notion of namespace.
It will be done in next API but while then, we tend to not be explicit about the namespace and use the "longUrl" as the url.

I don't think we should change the error message here.

MiguelRocha: I reverted the log msg back to what it was before

MiguelRocha:
> I think we could have two commits as it is two different changes.

I think the commit is small and its very easy to identify what is the out of bounds validation (I created a function for that) and what was the remove of duplicates on the links (I even inserted a comment there). But I will keep that in mind next time. Let me know if you agree with the other changes..

src/zimcheck.cpp
```diff
@@ -179,6 +179,27 @@ inline bool isInternalUrl(const std::string& input_string)
     return !isExternalUrl(input_string);
 }
 
+
+// checks if a relative path is out of bounds (relative to base)
+bool isOutofBounds(const std::string& input, const std::string& base)
+{
+    if (input.empty() || base.empty()) return false;
```
mgautierfr, line 186: An input `../../foo.html` on an empty `base` should return `true` no ?

mgautierfr:
> I think the commit is small and its very easy to identify what is the out of bounds validation (I created a function for that) and what was the remove of duplicates on the links (I even inserted a comment there).

The commits have to be identified without reading the change of the commits.
- The first line of the commit message (the title) must tell what is changed.
- The rest of the commit message (the description) must tell why this change.
  And sometime explain the how for complex commit (but maybe the commit need to be split then).
- The diff of the commit naturally tell the how.

More than a theoretical point, your two changes could are interesting enough to have a "why".
- The "out of bounds" check avoid a crash (That's the commit title says).
- The "remove duplicate links" doesn't just rewrite it correctly. It also sort the `links` vector. As we were not sorting the links before, the "previous link" was probably not equal to the link. So it simply fix the duplicate test, not just rewrite it.

Saying that, I realize that the duplication test was made on the "normalized link", not on the "original  link" (as written on the page). So the test is different, and probably wrong.
If the base is "A/foo/bar.html", two links "baz.html" and "../foo/baz.html" are different but resolve to the same normalized link.

On the out of bound part, I know that we don't have a unit test system in place. But it could be the good time to have it as I probably spot a bug in the `isOutOfBounds` function.

MiguelRocha:
@mgautierfr a separated PR is being written that introduces the unit tests. I've made changes to the code and commit history. Let me know your comments.

src/zimcheck.cpp
```diff
@@ -396,28 +396,35 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
...
+
+                filtered.push_back(normalize_link(l, baseUrl));
+            }
+
+            // remove duplicates
+            sort(filtered.begin(), filtered.end());
+            filtered.erase(unique(filtered.begin(), filtered.end()), filtered.end());
+            for(const auto &olink: filtered)
+            {
+                char nm = olink[0];
```
mgautierfr, line 441: The "o" of `olink` was probably here for "original link".
This is no more the original link as it is the normalized link.
`link` is a better name.

src/zimcheck.cpp
```diff
@@ -396,28 +396,35 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
...
-                        {
-                            std::ostringstream ss;
-                            ss << link << " (" << olink << ") was not found in article " << it->getLongUrl();
-                            reporter.addError(URL_INTERNAL, ss.str());
-                            previousLink = link;
-                            previousIndex = index;
-                        }
-                        reporter.setTestResult(URL_INTERNAL, false);
+                        std::ostringstream ss;
+                        ss << " (" << olink << ") was not found in article " << it->getLongUrl();
```
mgautierfr, line 450: We should display the link as it is written in the html. Else it may be difficult for the user to investigate and fix the issue.

MiguelRocha:
@mgautierfr I implemented you changes. Let me know what is the issue with the de duplication.

mgautierfr:
Two different original links can be resolved to the same normalized link.
The current implementation only keep one instance of the normalized link and may "delete" different wrong original links.

We should use a `pair<string, vector<string>>` mapping a normalized link to (potentially) different original links.

MiguelRocha:
@mgautierfr I implemented the changes related with the deduplication

mgautierfr:
We are good. Can you just rebase on `origin/master` ? Then we can merge.

MiguelRocha:
> We are good. Can you just rebase on `origin/master` ? Then we can merge.

Done

PR #131, opened 2020-07-21
Document new zimdump dependency in README

legoktm:
As of 07341975cc3, zimdump now requires the docopt C++ library.

Also minor grammar/copy editing fixes for the entire section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,15 @@ with the ZIM libary compilation itself, we recommend to have a look to
 Dependencies
 ------------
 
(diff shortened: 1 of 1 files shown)


legoktm:
Also Debian has a slightly older docopt version ([requested a newer one](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=965975)) without pkg-config support, so I had to build-depend upon `cmake` so meson could find docopt...is that expected?

Issue #98, opened 2020-04-20
Build a minimal CI and unit test for zimwriterfs


kelson42:
* Using gtest
* Github action
* Generating code coverage with codecove.io
* Using zimdump for checking output

kelson42:
@MiguelRocha @asashnov Please secure your or not doing duplicate work here. One of you needs to make basic CI working with gtest, then each of can work on its specific set of unit tests.

PR #142, opened 2020-08-04
Issue #98: add tests for src/tools.cpp

asashnov:
Related to #98 (partially fix)

* Add unittests for src/tools.cpp
* Remove some unused code
* Limit the scope of helper functions in src/tools.cpp


asashnov:
Seems CodeCov is not completely configured.

Crash of `CI / Linux (native_static)` in `commont-tools-test` may be caused by:
```
/usr/src/gtest/src/gtest.cc:3451: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

`CI / Linux (native_dyn)` fails to load libicu dynamic library:
```
/home/runner/zim-tools/build/test/commont-tools-test: error while loading shared libraries: libicui18n.so.58: cannot open shared object file: No such file or directory
```

I think those problems are environmental, not in the code.

test/commont-tools-test.cpp
```diff
@@ -0,0 +1,116 @@
+#include "gtest/gtest.h"
```
MiguelRocha, line 1: Typo in the filename

test/commont-tools-test.cpp
```diff
@@ -0,0 +1,116 @@
...
+  str = computeAbsolutePath("/home/alex/oz/zim-tools/test/data", "minimal-content/hello.html");
+  EXPECT_EQ(str, "/home/alex/oz/zim-tools/test/minimal-content/hello.html");
+}
+
+TEST(CommonTools, computeRelativePath)
+{
+  std::string str;
+
+  // FIXME: infinitive loop and memory bomb
+  // res = computeRelativePath("", "");
```
MiguelRocha, line 54: Avoid adding commented code. Since your testing this functions and discovering bugs, it is a good opportunity to fix them. We always want to increase the code quality.

test/meson.build
```diff
@@ -11,6 +12,15 @@ zimwriter_srcs = [  '../src/zimwriterfs/article.cpp',
                     '../src/tools.cpp']
 
 if gtest_dep.found() and not meson.is_cross_build()
+
+    test_tools_exe = executable('commont-tools-test',
```
MiguelRocha, line 16: Why is the icu-i18n lib necessary for the tools?

test/testhelpers.cpp
```diff
@@ -0,0 +1,11 @@
+#include <unistd.h>
+#include "testhelpers.h"
+#include "../src/tools.h"
+
+
+TestHelpers::TestHelpers()
```
MiguelRocha, line 6: I dont see the need for adding this class. Please use a helper function if nedded.

test/meson.build
asashnov: removeAccents() uses it:
```c
  static std::unique_ptr<icu::Transliterator> removeAccentsTrans(icu::Transliterator::createInstance(
      "Lower; NFD; [:M:] remove; NFC", UTRANS_FORWARD, status));
  icu::UnicodeString ustring(text.c_str());
  removeAccentsTrans->transliterate(ustring);
```

test/testhelpers.cpp
asashnov: Indeed. I thought I need more flexibility, but I didn't.

test/commont-tools-test.cpp
asashnov: Code fixed, more tests added.

test/commont-tools-test.cpp
asashnov: Renamed: commont-tools-test.cpp => tools-test.cpp

test/testhelpers.cpp
asashnov: class and files testhelpers.(cpp|h) removed.

test/tools-test.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "../src/tools.h"
+
+static std::string get_tests_data_dir()
+{
+  return "data";
+}
+
+TEST(CommonTools, getFileSize)
+{
+  std::string fn = get_tests_data_dir() + "/minimal-content/favicon.png";
```
mgautierfr, line 12: Wouldn't be better to create a file in tmp and test it is correctly detected as existing instead of storing a file in the repository ?

test/tools-test.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+  unsigned char data[] = { 0xff, 0x00, 0x7a };
+  std::string txt = base64_encode(data, sizeof(data));
+  EXPECT_EQ(txt, "/wB6");
+}
+
+TEST(CommonTools, decodeUrl)
+{
+  const char *src = "https://www.example.com/cgi-bin/search.cgi?q=example%20search";
+  std::string res = decodeUrl(src);
```
mgautierfr, line 28: We should test more decoding char than `%20` only.

test/tools-test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  replaceStringInPlace(str, "", "");
+  EXPECT_EQ(str, "");
+
+  str = "abcd";
+  replaceStringInPlace(str, "a", "b");
+  EXPECT_EQ(str, "bbcd");
+
+  str = "abcd";
+  replaceStringInPlace(str, "a", "");
+  EXPECT_EQ(str, "bcd");
```
mgautierfr, line 107: We may add this test in `replaceStringInPlaceOnce` test also.

test/meson.build
```diff
@@ -1,4 +1,5 @@
 gtest_dep = dependency('gtest', main:true, fallback:['gtest', 'gtest_main_dep'], required:false)
+icu_dep = dependency('icu-i18n', static:static_linkage)
```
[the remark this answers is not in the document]
asashnov: I looked carefully through https://mesonbuild.com/Dependencies.html  and didn't found that a dependency can be associated with a source file. Dependencies always specified to the exe file only.

libicu is required by libzim, so maybe we should just link our tests executable with libzim, as we do it in tools directory:

```
executable('zimpatch', 'zimpatch.cpp',
  dependencies: libzim_dep,
  install: true)
```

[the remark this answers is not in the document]
asashnov: Extra favicon.png is removed (from 'data/...', leaved in 'test/data/..')

test/tools-test.cpp
```diff
@@ -0,0 +1,141 @@
+#include "gtest/gtest.h"
+
+#include "../src/tools.h"
+
+static std::string get_tests_data_dir()
+{
+  return "data";
+}
```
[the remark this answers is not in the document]
asashnov: I wanted to leave it for future flexibility, but indeed, it's better to keep things as simple as it can be. Removed.

test/tools-test.cpp
asashnov: Probably, no, as this will require extra lines of code. And favicon.png is used anyway in further tests (#143)

test/meson.build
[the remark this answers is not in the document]
mgautierfr: icu is used in tools (https://github.com/openzim/zim-tools/blob/master/src/tools.cpp#L302) so it is not a requirement of libzim (in fact, it is but we cannot rely on this).
All binaries that use `removeAccents` need to link with icu. Not only the tests.
So we need to move the icu dependency global and make "real" binaries linking with it.

On the tests side, it is simpler to make all test linking with it (as it is done with other deps).

For now binaries are working because libzim also depend on it (if compiled with xapian). But we should not rely on this side effect.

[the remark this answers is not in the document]
asashnov: Actually, we don't need libicu and 'removeAccents' function in zim-tools at all.
Seems it was unused copy/paste from libzim. I just removed `removeAccents()` code and test.
Now we don't need a special case for 'tools-test'.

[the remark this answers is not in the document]
asashnov: `removeAccents()` seems to be unused copy/paste from libzim. So I just removed it.

src/tools.cpp
```diff
@@ -294,20 +295,6 @@ std::string getNamespaceForMimeType(const std::string& mimeType, bool uniqueName
   }
 }
 
-
-std::string removeAccents(const std::string& text)
```
mgautierfr, line 298: We must remove the declaration also.

kelson42:
@MiguelRocha Good for you as well?

src/tools.cpp
```diff
@@ -254,6 +252,9 @@ void replaceStringInPlace(std::string& subject,
                           const std::string& search,
                           const std::string& replace)
 {
+  if (search.empty())
```
MiguelRocha, line 255: If the search is empty if means that we replace nothing. It should return the original string here.

test/tools-test.cpp
```diff
@@ -0,0 +1,196 @@
...
+std::string directoryPath;
+magic_t magic;
+bool inflateHtmlFlag = false;
+bool isVerbose() { return false; }
+
+TEST(CommonTools, getFileSize)
+{
+  std::string fn = "data/minimal-content/favicon.png";
+
+  system("pwd");
```
MiguelRocha, line 15: this system() syscall was probably for debugging?

test/tools-test.cpp
```diff
@@ -0,0 +1,196 @@
...
+  EXPECT_EQ(res[0], '\0');
+
+  src = "%ff";
+  res = decodeUrl(src);
+  EXPECT_EQ(res.size(), 1);
+  EXPECT_EQ(res[0], '\xff');
+
+
+  // test corner cases
+  src = "";
```
MiguelRocha, line 51: Here we could have a map of (original string, expected string) and then run it on a loop.

_pseudo code_
```
unordered_map<string, string> = expectationsMap = {  {"%", "%"}, {"%2", "%2"}, ..., ... };

for (auto p : expectationsMap) EXCPECT_EQ(p.first, p.second);

```

But its just a suggestion. No need to implement it for approval.

(I later noticed that there are other places in the code that this could be done)

test/tools-test.cpp
asashnov: Yes, it was for debug, removed. Excuse me.

test/tools-test.cpp
asashnov: Good idea to have at least one example in the code. One test was rewritten.

src/tools.cpp
MiguelRocha: True. Sorry for the confusion.

MiguelRocha (approved):

PR #144, opened 2020-08-08
Adds unit testing for tools

MiguelRocha:
Moves utility code from zimcheck to tools. Adds unit testing to tools

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,10 +6,10 @@ coverage:
   status:
     project:
       default:
-        threshold: 1%
+        threshold: 5%
     patch:
       default:
-        target: 90%
+        target: 80%
         threshold: 0%
 
 ignore:
--- a/src/meson.build
+++ b/src/meson.build
@@ -26,7 +26,7 @@ executable('zimsplit', 'zimsplit.cpp',
   dependencies: libzim_dep,
   install: true)
 
-executable('zimcheck', 'zimcheck.cpp',
+executable('zimcheck', 'zimcheck.cpp', 'tools.cpp',
   dependencies: libzim_dep,
   install: true)
 
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -31,6 +31,7 @@
 #include <vector>
 #include <memory>
 #include <unistd.h>
+#include <algorithm>
 
 #ifdef _WIN32
 #define SEPARATOR "\\"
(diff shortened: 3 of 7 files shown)


test/test-tools.cpp
```diff
@@ -0,0 +1,65 @@
...
+    ASSERT_EQ(normalize_link("a", "/b"), "/b/a");
+    ASSERT_EQ(normalize_link("../a", "/b/c"), "/b/a");
+    ASSERT_EQ(normalize_link(".././a", "/b/c"), "/b/a");
+    ASSERT_EQ(normalize_link("../a/b/aa#localanchor", "/b/c"), "/b/a/b/aa");
+    ASSERT_EQ(normalize_link("../a/b/aa?localanchor", "/b/c"), "/b/a/b/aa");
+}
+
+TEST(tools, addler32)
+{
+    ASSERT_EQ(adler32("sdfkhewruhwe8"), 640746832);
```
mgautierfr, line 49: Test with at least two different values (and a empty string).

`int adler32(std::string buf) { return 640746832; }` is a implementation passing this test.

test/test-tools.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    ASSERT_TRUE(v.empty());
+
+    std::string page1 = "<link href=\"https://fonts.goos.com/css?family=OpenSans\" rel=\"stylesheet\">";
+    auto v1 = generic_getLinks(page1);
+
+    ASSERT_FALSE(v1.empty());
+    ASSERT_TRUE(v1.size() == 1);
+    ASSERT_EQ(v1[0], "https://fonts.goos.com/css?family=OpenSans");
+}
```
mgautierfr, line 64: `generic_getLinks` find links with `src="foo"`, we should test this.
It also get an argument `withHref` to skip or not `href=""` links.
We should test that it is correctly taken into account.

test/test-tools.cpp
```diff
@@ -0,0 +1,65 @@
...
+  ~ToolsTest() {
+      magic_close(magic);
+  }
+};
+
+TEST(tools, isOutofBounds)
+{
+    ASSERT_FALSE(isOutofBounds("", ""));
+    ASSERT_TRUE(isOutofBounds("../../..", ""));
+    ASSERT_TRUE(isOutofBounds("../", ""));
```
mgautierfr, line 32: Mayby a test (or several) with a not empty base also ?

mgautierfr:
Final code is good.
But you have split the commit in two (one to move code in tools, one to add unit test).
Then you need to update the commit message (because the fist one says that it also add unit test).

mgautierfr (approved):

PR #147, opened 2020-08-14
Remove useless log entry telling its OK

kelson42:
This is the only place where `zimcheck` was doing that. We should be coherent.


mgautierfr:
@kelson42 Please merge PR with the default github actions (`Create a merge commit`) and don's use the others (`Squash and merge` or `Rebase and merge`).

PR #148, opened 2020-08-16
Detect empty links and report accordingly

MiguelRocha:
Reports a more appropriate message when zimcheck finds empty links inside an article.

### DIFF
--- a/src/zimcheck.cpp
+++ b/src/zimcheck.cpp
@@ -295,10 +295,16 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
 
             auto links = generic_getLinks(it->getData());
             std::unordered_map<std::string, std::vector<std::string>> filtered;
+            int nremptylinks = 0;
             for (const auto &l : links)
             {
                 if (l.front() == '#' || l.front() == '?') continue;
                 if (isInternalUrl(l) == false) continue;
+                if (l.empty())
+                {
+                    nremptylinks++;
+                    continue;
+                }
 
                 if (isOutofBounds(l, baseUrl))
                 {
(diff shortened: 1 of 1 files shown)


MiguelRocha:
@kelson42 output:

```
...
 full_by_title.js (idx 70) and auth_41984_lang_ar_by_title.js (idx 3)
 lang_ar_by_title.js (idx 85) and auth_41984_lang_ar_by_title.js (idx 3)
  Found 2 of empty links in article: A/Home.html
[INFO] Overall Test Status: Fail

```

kelson42:
@MiguelRocha Thank you, output looks good to me.

kelson42:
@MiguelRocha A unit test would be welcome here I believe

mgautierfr:
> A unit test would be welcome here I believe

Yes, but it is complicated to test is right now as there is not free function to test.

It would be nice to move the code checking for the article content in one (or several) function(s) working only on the content and not have one function mixing looping hover the articles and checking them.

But I'm not sure we should have this change in this PR.

MiguelRocha:
@mgautierfr agree. Also after https://github.com/openzim/zim-tools/pull/144 is merged we can startup pushing the unit tests very easly :)

src/zimcheck.cpp
```diff
@@ -437,23 +437,38 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
...
+                    nremptylinks++;
+                    continue;
+                }
 
                 if (isOutofBounds(l, baseUrl))
                 {
                     std::ostringstream ss;
                     ss << l << " is out of bounds. Article: " << it->getLongUrl();
                     reporter.addError(TestType::URL_INTERNAL, ss.str());
+                    reporter.setTestResult(TestType::URL_INTERNAL, false);
```
mgautierfr, line 456: Is this really related to empty links detection ?

If not we should move it to a different commit.

mgautierfr (changes requested):

src/zimcheck.cpp
MiguelRocha: Done

mgautierfr (approved):

MiguelRocha:
@kelson42 I think we can merge this one. I'm also working on the full automated code for zimcheck in this PR: https://github.com/openzim/zim-tools/pull/153/files

Issue #100, opened 2018-09-18
zimwriterfs crashes after attempting to read dead symlink


MAGNUS-0:
# OS & Versions 

**zimwriterfs**
```
$ zimwriterfs -V
Version:  v1.2
	See -h or --help argument flags for further argument options
	Copyright 2013-2016 Emmanuel Engelhart <kelson@kiwix.org>
```

**OS**
```
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
```

# Command being run
File : **zimsco.sh**
```
zimwriterfs \
--welcome="./Vowel.html" \
--favicon="./favicon.png" \
--language=eng \
--title="Wikisco" \
--description="Scottish Wiki" \
--creator="Wikipedia" \
--publisher="Kiwix" \
./wikipedia_sco_all_2018-09 \
./wikisco_output | tee ./wikisco_output
```
Command : 
`$ time sudo ./zimsco.sh`

# Console output
```
A:1000; CA:999; UA:1; FA:0; IA:991; C:6; CC:6; UC:0
A:2000; CA:1998; UA:2; FA:1; IA:1990; C:12; CC:12; UC:0
A:3000; CA:2998; UA:2; FA:1; IA:2990; C:19; CC:19; UC:0
A:4000; CA:3963; UA:37; FA:39; IA:3935; C:26; CC:26; UC:0
A:5000; CA:4963; UA:37; FA:39; IA:4935; C:33; CC:33; UC:0
A:6000; CA:5963; UA:37; FA:39; IA:5935; C:39; CC:39; UC:0
A:7000; CA:6963; UA:37; FA:39; IA:6935; C:46; CC:46; UC:0
A:8000; CA:7963; UA:37; FA:39; IA:7935; C:52; CC:52; UC:0
A:9000; CA:8963; UA:37; FA:53; IA:8921; C:58; CC:58; UC:0
A:10000; CA:9963; UA:37; FA:53; IA:9921; C:65; CC:65; UC:0
A:11000; CA:10963; UA:37; FA:53; IA:10921; C:72; CC:72; UC:0
A:12000; CA:11963; UA:37; FA:53; IA:11921; C:78; CC:78; UC:0
A:13000; CA:12963; UA:37; FA:53; IA:12921; C:85; CC:85; UC:0
A:14000; CA:13963; UA:37; FA:53; IA:13921; C:92; CC:92; UC:0
[...]

PR #158, opened 2020-09-09
Issue #102: create ZIM redirect entry for file symlinks.

asashnov:
Issue #102: create ZIM redirect entry for file symlinks.

Skip broken, cycled or pointing outside of HTML directory symlinks.

### DIFF
--- a/src/tools.h
+++ b/src/tools.h
@@ -24,6 +24,36 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <stdexcept>
+#include <sstream>
+
+/* Formatter for std::exception what() message:
+ * throw std::runtime_error(
+ *   Formatter() << "zimwriterfs: Unable to read" << filename << ": " << strerror(errno));
+ */
+class Formatter
+{
+public:
+    Formatter() {}
+    ~Formatter() {}
+
+    template <typename Type>
+    Formatter & operator << (const Type & value)
+    {
+        stream_ << value;
+        return *this;
+    }
+
+    // std::string str() const         { return stream_.str(); }
+    operator std::string () const   { return stream_.str(); }
+
+private:
+    Formatter(const Formatter &);
+    Formatter & operator = (Formatter &);
+
+    std::stringstream stream_;
+};
+
 
 typedef struct html_link
 {
@@ -39,6 +69,7 @@ std::string decodeUrl(const std::string& encodedUrl);
 std::string computeAbsolutePath(const std::string& path,
                                 const std::string& relativePath);
 bool fileExists(const std::string& path);
+bool isDirectory(const std::string &path);
 
 std::string base64_encode(unsigned char const* bytes_to_encode,
                           unsigned int in_len);
--- a/src/zimwriterfs/zimcreatorfs.cpp
+++ b/src/zimwriterfs/zimcreatorfs.cpp
@@ -26,8 +26,29 @@
 #include <dirent.h>
 #include <sys/stat.h>
 #include <regex>
+#include <unistd.h>
+#include <limits.h>
+#include <cassert>
 
 bool isVerbose();
+
+ZimCreatorFS::ZimCreatorFS(std::string _directoryPath, std::string mainPage,
+                           bool verbose, bool uniqueNamespace, bool zstd)
+  : zim::writer::Creator(verbose, zstd ? zim::zimcompZstd : zim::zimcompLzma),
+    directoryPath(_directoryPath),
+    mainPage(mainPage),
+    uniqueNamespace(uniqueNamespace)
+{
+  char buf[PATH_MAX];
+
+  if (realpath(directoryPath.c_str(), buf) != buf) {
+    throw std::invalid_argument(
+          Formatter() << "Unable to canonicalize HTML directory path "
+                      << directoryPath << ": " << strerror(errno));
+  }
+
+  canonical_basedir = buf;
+}
 
 zim::writer::Url ZimCreatorFS::getMainUrl() const
 {
@@ -91,10 +112,10 @@ void ZimCreatorFS::visitDirectory(const std::string& path)
 
     switch (entry->d_type) {
       case DT_REG:
+        addArticle(fullEntryName);
+        break;
       case DT_LNK:
-        {
-          addArticle(fullEntryName);
-        }
+        processSymlink(path, fullEntryName);
         break;
       case DT_DIR:
         visitDirectory(fullEntryName);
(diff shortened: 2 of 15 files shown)


asashnov:
Example of the output:

```
$ /home/alex/oz/zim-tools/build/src/zimwriterfs/zimwriterfs \
>     -w index.html \
>     --uniqueNamespace -f icon.png -l EN -t "Testing Heuristics" \
>     -d "Early cut of the testing heuristics material." -c "Julian Harty" \
>     -p "Commercetest Ltd." testing-heuristics/dist th-ln.zim
Unable to resolve symlink testing-heuristics/dist/A/style.css: Too many levels of symbolic links
```
(this is for self-pointing symlink, created as the ticket says. No crash now.)

Output of unit test with 3 kinds of links:
```
Skip symlink data/with-symlink/symlink-outside.html: points outside of HTML directory
Unable to resolve symlink data/with-symlink/symlink-self.html: Too many levels of symbolic links
Unable to resolve symlink data/with-symlink/symlink-not-existing.html: No such file or directory
```

kelson42:
@asashnov I believe this PR should as well fix #100 as this is strongly related I believe.

test/zimwriterfs-zimcreatorfs.cpp
```diff
@@ -75,18 +75,63 @@ TEST(ZimCreatorFSTest, MinimalZim)
...
 }
+
+TEST(ZimCreatorFSTest, SymlinkShouldCreateRedirectArticleEntry)
+{
+  LibMagicInit libmagic;
+
+  std::string directoryPath = "data/with-symlink";
+  ZimCreatorFS zimCreator(directoryPath, "hello.html", false, false);
+
+  TempFile out("with-symlink.zim");
```
mgautierfr, line 107: Do we really want to create a temporaryfile just to get it path ?

test/zimwriterfs-zimcreatorfs.cpp
```diff
@@ -75,18 +75,63 @@ TEST(ZimCreatorFSTest, MinimalZim)
...
+
+  zimCreator.startZimCreation(out.path());
+  zimCreator.visitDirectory(directoryPath);
+
+  zimCreator.addArticle(
+        std::make_shared<MetadataFaviconArticle>(zim::writer::Url('I', "link.jpeg")));
```
mgautierfr, line 113: Why do we need this article ?

asashnov (commented):
TempFile also deletes the file in its destructor.

test/zimwriterfs-zimcreatorfs.cpp
asashnov: It's 'favicon' metadata which usually added in the same way from main() in zimwriterfs.cpp
It's equivalent of calling `zimwriterfs -f link.jpeg`:

Screenshot from 2020-09-15 12-43-27

test/zimwriterfs-zimcreatorfs.cpp
```diff
@@ -75,18 +75,63 @@ TEST(ZimCreatorFSTest, MinimalZim)
...
+  // VERIFY the created .zim file with 'zimdump'
   zim::File zimfile(out.path());
   EXPECT_EQ(zimfile.getCountArticles(), 4);
 
   zim::Article a1 = zimfile.getArticle('A', "index.html");
   EXPECT_TRUE(a1.isRedirect());
 
+  uint32_t main_index = zimfile.getFileheader().getMainPage();
+  EXPECT_EQ(a1.getIndex(), main_index);
+
```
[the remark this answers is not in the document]
asashnov: Will create another PR

src/zimwriterfs/zimcreatorfs.cpp
```diff
@@ -164,6 +169,66 @@ void ZimCreatorFS::addArticle(std::shared_ptr<zim::writer::Article> article)
...
 
+void ZimCreatorFS::processSymlink(const std::string& curdir, const std::string& symlink_path)
+{
+  /* #102 Links can be 3 different types:
+   *  - dandling (not pointing to a valid file)
+   *  - pointing to file but outside of 'directoryPath'
+   *  - looped symlinks
+   */
+  char buf[PATH_MAX];
+  memset(buf, 0, sizeof(buf)); // the manpage says readlink() won't null terminate
```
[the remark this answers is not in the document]
asashnov: readlink() isn't used anymore.

src/zimwriterfs/zimcreatorfs.cpp
```diff
@@ -26,11 +26,16 @@
...
+#include <unistd.h>
+#include <limits.h>
+#include <cassert>
 
 bool isVerbose();
 
 zim::writer::Url ZimCreatorFS::getMainUrl() const
 {
+  // The index of main page will be stored in ZIM file header
+  // (not as redirect, as it's done for favicon)
```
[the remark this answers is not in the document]
mgautierfr: > Why favicon is linked, and mainpage have just its number stored?

For historical reason. mainpage is stored in the zim file header. We add the favicon without changing the zim format and so we add it as a "well known path" (a redirect to the real favicon).

Note that the new writer api also change that a bit. The creator will create a redirect for the main page (as the favicon) and store the id of the redirect in the zim header.

test/zimwriterfs-zimcreatorfs.cpp
mgautierfr: Yes, but we don't need it. We can create a zim without the favicon. And we don't test it.
There is no need to create a zim file with it (and store the file in the git history).

asashnov (commented):
Rebased on top of the current 'master', link.jpeg removed.

src/zimwriterfs/zimcreatorfs.cpp
```diff
@@ -26,9 +26,32 @@
...
+    directoryPath(_directoryPath),
+    mainPage(mainPage),
+    uniqueNamespace(uniqueNamespace)
+{
+  char buf[PATH_MAX];
+
+  if (realpath(directoryPath.c_str(), buf) != buf) {
+    std::cerr << "Unable to canonicalize HTML directory path " << directoryPath
+              << ": " << strerror(errno) << std::endl;
+    return;
```
mgautierfr, line 48: I try to avoid possibilities of creating invalid objects.
If the arguments of the constructor of `ZimCreatorFS` are not valid, we must not create a `ZimCreatorFS`.

It would be better to raise an exception here.

src/zimwriterfs/zimwriterfs.cpp
```diff
@@ -338,6 +338,9 @@ int main(int argc, char** argv)
 
   ZimCreatorFS zimCreator(directoryPath, welcome, isVerbose(), uniqueNamespace, zstdFlag);
 
+  if (! zimCreator.isInitSuccess())
+    exit(1);
+
```
mgautierfr, line 343: This is useless if we cannot create a invalid creator.

mgautierfr (changes requested):
It would be better to raise a exception instead of constructing a invalid ZimCreatorFS.

You may be pretty sure that user code will not check for the init success and this will fail. If we raise a exception, the code will stop as soon as we have detect the error.

src/zimwriterfs/zimcreatorfs.cpp
asashnov: Done, an exception is raised.

src/tools.h
```diff
@@ -24,6 +24,36 @@
...
 #include <string>
 #include <vector>
+#include <stdexcept>
+#include <sstream>
+
+/* Formatter for std::exception what() message:
+ * throw std::runtime_error(
+ *   Formatter() << "zimwriterfs: Unable to read" << filename << ": " << strerror(errno));
+ */
+class Formatter
```
mgautierfr, line 34: This class is nice. I may worth to move this in libzim's tools and use it everywhere we create a stringstream in our code.
This could even be merged with a `to_string` function.

mgautierfr (approved):

Issue #134, opened 2020-07-29
zimcheck performance improvement


MiguelRocha:
zimcheck performance is quite important since for big articles can take several minutes (maybe hours).

There are a few performance improvements that can be done in zimcheck specific code:

- Decrease the number of function calls to `getLinks()`
- Improve `getLinks()` performance
- Improve `isExternalUrl()`

The first 2 points are responsible for 18% while the last one takes 20% speedtime.

Also, there maybe some performance gains if we find a better way to validate the URLs. Currently we are calling getArticle() from libzim and this takes around 50% of the time since it needs access the direntCache. (related: https://github.com/openzim/libzim/issues/385#issue-667665866)

kelson42:
@MiguelRocha Good to me to implement. If the amount of changes is subtantial, please split this ticket.

veloman-yunkan:
@MiguelRocha Do you mind if I work on this in combination with optimizations on the libzim side?

MiguelRocha:
@veloman-yunkan please feel free to do it. And let me know if you need any help. I have a draft branch that already fixes the first point that I described and I will create a PR after finishing other issues.

kelson42:
@MiguelRocha Would you be able to move the `isExternalUrl()` in an other ticket please
@veloman-yunkan Thank you for your offer, I propose @MiguelRocha make a PR with what he already has done and then let discuss this ticket again.

kelson42:
@mgautierfr @MiguelRocha I believe this should not be closed because the following points need to be implemented:
* Improve `getLinks()` performance
* Improve `isExternalUrl()`

mgautierfr:
Yes, it was automatically closed when I've merged #162.

kelson42:
There is no obvious and consensual solution to optimize `getLinks()` and `isExternalUrl()`. Closing for the moment and we will open a new ticket if a clear (speed?) problem happens again.

PR #162, opened 2020-09-20
Reduced the number of calls to generic_getLinks.

MiguelRocha:
Refactored generic_getLinks to parse all src and href type links.
Removed getDependencies()
Make zimcheck call generic_getLinks only once

### DIFF
--- a/src/tools.h
+++ b/src/tools.h
@@ -25,6 +25,12 @@
 #include <string>
 #include <vector>
 
+typedef struct html_link
+{
+    std::string attribute;
+    std::string link;
+} html_link;
+
 std::string getMimeTypeForFile(const std::string& basedir, const std::string& filename);
 std::string getNamespaceForMimeType(const std::string& mimeType, bool uniqueNamespace);
 std::string getFileContent(const std::string& path);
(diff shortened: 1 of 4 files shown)


kelson42:
@MiguelRocha Do we have a ticket for that? Is that a speed-up PR?

src/zimcheck.cpp
```diff
@@ -361,13 +356,14 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
             if (it->getMimeType() != "text/html")
                 continue;
 
-            auto links = getDependencies(it->getPage());
-            for (auto &link: links)
+            if (url_check == false)
+                links = generic_getLinks(it->getPage());
```
mgautierfr, line 360: Maybe move this in line 289 with :
```
if (url_check || url_check_external) {
    links = generic_getLinks(it->getData())
}
```
(`getPage` will be removed in next api)

src/tools.h
```diff
@@ -25,6 +25,12 @@
 #include <string>
 #include <vector>
 
+typedef struct html_link
+{
+    std::string attribute;
+    std::string link;
+} html_link;
+
```
mgautierfr, line 33: I have to say that I would have naively use a `std::tuple` or a `std::pair`, but it is better with a specific structure.

mgautierfr (changes requested):
Just a small change about where we initialize `links`.
But else we are ok.

src/zimcheck.cpp
MiguelRocha: done

src/tools.h
MiguelRocha: I was thinking about that as well but I wanted to be more expressive here to avoid confusion.

MiguelRocha:
> @MiguelRocha Do we have a ticket for that? Is that a speed-up PR?

Yes it is the first point of https://github.com/openzim/zim-tools/issues/134

mgautierfr (approved):

PR #163, opened 2020-10-03
Remove arg zimsplit

MiguelRocha:
This PR adds docopt command line parsing to zimsplit and removes the need for arg.h.


mgautierfr (changes requested):
Code is good.
However, prefix and size arguments are mandatory. We were not ensuring that is previous version but if we change argument parsing it may be good to refuse empty `prefix` or `size`.
Or use useful default values (prefix=input_zim_file and size=2Gb).

veloman-yunkan (approved):

MiguelRocha:
@mgautierfr I added a default prefix and for the size I turned it mandatory.

src/zimsplit.cpp
```diff
@@ -159,40 +163,43 @@ class ZimSplitter
...
-        << std::flush;
-      return -1;
-    }
+    std::string versionstr("zimsplit " + std::string(VERSION));
+    std::map<std::string, docopt::value> args = docopt::docopt(USAGE,
+                                                            {argv + 1, argv + argc},
+                                                            true,
+                                                            versionstr);
+
+    std::string prefix = PARTPREFIX;
```
mgautierfr, line 193: I would use the file input as default prefix.
This way, a `zimsplit foo.zim` will generate `foo.zimaa`, `foo.zimab`, ...

mgautierfr (changes requested):
I think after all that we should have a default value for the size (2GB).
We used to split zim using this size when we were using simple cut tool.
And the size as to be specified in Bytes (2*1024*1024*1024=2147483648), this is inconvenient enough to provide a default value.

src/zimsplit.cpp
```diff
@@ -159,40 +162,45 @@ class ZimSplitter
...
+                                                            true,
+                                                            versionstr);
+
+    std::string prefix = args["<file>"].asString();
+    if (args["--prefix"])
+        prefix = args["--prefix"].asString();
+
+    int size =  DEFAULT_PART_SIZE;
+    if (args["--size"])
+	size = args["--size"].asLong();
```
mgautierfr, line 198: Use space not tab.

mgautierfr (approved):

mgautierfr:
I'm ok with the PR.
As for #169 I'm waiting for the CI to be ok to be sure.

There is this cosmetic issue with the tab. You may want to fix while we are waiting but I will not block for this.

src/zimsplit.cpp
kelson42: done

kelson42: done

PR #169, opened 2020-10-27
Removed redundant call to generic_getLinks()

veloman-yunkan:
A redundant call to `generic_getLinks()` was introduced by commit a46e5d93 "created zimcheck src directory".


veloman-yunkan:
I wonder if the failure of the Linux CI test runs is due to openzim/libzim#430

MiguelRocha (approved):

veloman-yunkan:
It is now fixed by openzim/libzim#440

mgautierfr (approved):
Nice catch.

I'm just waiting that nightly CI build correct libzim dependency and re-run the CI for this PR. Just to be sure that we don't have a hidden bug we don't know about.

PR #181, opened 2020-11-02
Handling of root url (/) links in zimcheck

### LINKED ISSUE #180: New zimcheck crash scenario
With cutting edge libzim/zimtools:
```
$ zimcheck archlinux_en_all_maxi_2020-11.zim 
[INFO] Checking zim file archlinux_en_all_maxi_2020-11.zim
[INFO] Verifying Internal Checksum...
[INFO] Searching for metadata entries...
[INFO] Searching for main page...
[INFO] Verifying Articles' content...
basic_string::substr: __pos (which is 2) > this->size() (which is 0)
```

Source ZIM file: https://mirror.download.kiwix.org/zim/other/archlinux_en_all_maxi_2020-11.zim

veloman-yunkan:
Fixes #180 

An absolute link to the root url (e.g. href="/") cannot be decomposed to a namespace and path components. Luckily, `zim::File` has a `getArticleByUrl(const std::string& url)` method which handles that.

### DIFF
--- a/src/zimcheck/checks.cpp
+++ b/src/zimcheck/checks.cpp
@@ -180,10 +180,8 @@ void test_articles(const zim::File& f, ErrorLogger& reporter, ProgressBar progre
 
             for(const auto &p: filtered)
             {
-                std::string link = p.first;
-                char nm = link[0];
-                std::string shortUrl(link.substr(2));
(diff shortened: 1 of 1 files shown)


mgautierfr:
Good.
This also goes in the direction of libzim next api where namespace is not part of the api.

PR #183, opened 2020-11-03
Fix the type of DEFAULT_PART_SIZE 

mgautierfr:
2147483648 is bigger for a signed 32bits and is cast to -2147483648.

As we are setting the right type, lets do it everywhere and use
`zim::size_type` where it makes sense.


kelson42:
@mgautierfr I guess this is ready to review?

veloman-yunkan (approved):

PR #196, opened 2020-11-24
Adapt zim-tools to the new api of libzim (libzim_next).

mgautierfr:
Need openzim/libzim#454 first.

First commits should be relatively easy to review (although a bit long).

Port of zimwriterfs is a bit more complex and has been split in several (last) commits.
Those commits are split to simplify review but they should be taken all together to have something coherent.


src/zimpatch.cpp
```diff
@@ -39,265 +41,158 @@ std::string NumberToString(int number)
...
+      zimCreator.addRedirection(entry.getPath(), entry.getTitle(), redirectPath);
+      continue;
+    } catch(...) {
+      if (entry.isRedirect()) {
+        zimCreator.addRedirection(entry.getPath(), entry.getTitle(), entry.getRedirectEntry().getPath());
+      } else {
+        auto tmpItem = std::shared_ptr<zim::writer::Item>(new CopyItem(entry.getItem()));
+        zimCreator.addItem(tmpItem);
+      }
     }
```
veloman-yunkan, line 191: Code duplication with a corresponding block in the previous loop. Looks like it can be converted into a free function taking only two parameters: `addEntry(zimCreator, entry)`

veloman-yunkan (changes requested):
This is the 1st part of the review. I looked over the last few commits very quickly and will revisit them tomorrow.

src/zimwriterfs/zimcreatorfs.h
```diff
@@ -29,8 +29,9 @@
 class IHandler
 {
  public:
-  virtual void handleArticle(std::shared_ptr<zim::writer::Article> article) = 0;
-  virtual std::shared_ptr<zim::writer::Article> getMetaArticle() = 0;
+  virtual void handleItem(std::shared_ptr<zim::writer::Item> item) = 0;
```
veloman-yunkan, line 32: Judging from the commit "Adapt mimetypecounter." (and assuming that it didn't break compilation) the method `IHandler::handleItem()` isn't called (and neither was the old method `IHandler::handleArticle()` called). UPDATE: The answer to this conjecture was found in the previous ("Correctly add file item.") commit (see the comment therein).

Other than that, don't you think that passing the item by a const reference would be more appropriate? Do you envision a handler that will need to store an item passed via `handleItem()` until `getData()` is called.

src/zimwriterfs/zimcreatorfs.cpp
```diff
@@ -159,20 +159,41 @@ void ZimCreatorFS::visitDirectory(const std::string& path)
   closedir(directory);
 }
 
-void ZimCreatorFS::addArticle(const std::string& path)
+void ZimCreatorFS::addFile(const std::string& path)
 {
-  auto farticle = std::make_shared<FileArticle>(this, path);
```
veloman-yunkan, line 173: farticle - I feel sad that such a funny identifier is removed from the source code

src/zimwriterfs/zimcreatorfs.cpp
```diff
@@ -159,20 +159,41 @@ void ZimCreatorFS::visitDirectory(const std::string& path)
...
+void ZimCreatorFS::addItem(std::shared_ptr<zim::writer::Item> item)
 {
-  Creator::addArticle(article);
-  for (auto& handler: articleHandlers) {
-      handler->handleArticle(article);
+ Creator::addItem(item);
+ for (auto& handler: itemHandlers) {
+     handler->handleItem(item);
   }
 }
```
veloman-yunkan, line 187: This change must belong to the next commit "Adapt mimetypecounter." (which renames `articleHandlers` to `itemHandlers` and changes the IHandler interface).

veloman-yunkan (changes requested):
The second and last part of the review. The review of such a large change is far from being thorough, but I tried to do my (reasonable) best.

src/zimcheck/checks.cpp
```diff
@@ -65,93 +66,98 @@ void test_metadata(const zim::File& f, ErrorLogger& reporter) {
...
+        if (entry.isRedirect() || ns == 'M') {
+            continue;
         }
 
-        if (it->isRedirect() ||
-            it->isLinktarget() ||
-            it->isDeleted() ||
-            it->getArticleSize() == 0 ||
-            it->getNamespace() == 'M') {
+        if (empty_check & (ns=='A' || ns == 'I')) {
```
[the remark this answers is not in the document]
mgautierfr: Nice catch

src/zimcheck/checks.cpp
```diff
@@ -65,93 +66,98 @@ void test_metadata(const zim::File& f, ErrorLogger& reporter) {
...
         std::vector<html_link> links;
         if (url_check || url_check_external) {
-            links = generic_getLinks(it->getData());
+            links = generic_getLinks(data);
         }
 
         if(url_check)
         {
-            auto baseUrl = it->getLongUrl();
+            auto baseUrl = item.getPath();
```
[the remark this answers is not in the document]
mgautierfr: Yes, the same few lines above.

src/zimdump.cpp
```diff
@@ -111,195 +113,117 @@ std::string urlEncode(const std::string& value, bool encodeReserved)
...
-  if (m_file.getFileheader().hasChecksum()) {
-    std::cout <<
-               "checksum pos: " << m_file.getFileheader().getChecksumPos() << "\n"
-               "checksum: " << m_file.getChecksum() << "\n";
+  std::cout << "count-entries: " << m_archive.getEntryCount() << "\n";
+  std::cout << "uuid: " << m_archive.getUuid() << "\n"
+            <<  "cluster count: " << m_archive.getClusterCount() << "\n";
+  if (m_archive.hasChecksum()) {
+    std::cout << "checksum: " << m_archive.getChecksum() << "\n";
   } else {
```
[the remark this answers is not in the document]
mgautierfr: For me yes, obviously :)

The different offsets of the internal arrays are pretty useless out of libzim. There is absolutely no interest to have them at `zimdump` level.
If someone want to explore a zim file, it will need to parse the binary, it can parse the header.

For the namespace, one main goal of this new api is to remove them, so it make sens also.

src/zimpatch.cpp
mgautierfr: Yes. But I really don't want to spend to much time on zimpatch/zimdiff.
They are tools created on long time ago and I disagree with the code itself.

I agree with the idea of having this kind of tools but I think this should be tackled at an another level. Something lower level working directly on cluster instead of article. And using another format that zim to store the patch file.

The code is here and we have to port it to new libzim api but probably nothing more (at least for now).

src/zimwriterfs/zimcreatorfs.h
mgautierfr: > Other than that, don't you think that passing the item by a const reference would be more appropriate? Do you envision a handler that will need to store an item passed via handleItem() until getData() is called.

It was a time where the ArticleHandler was also use to index the article. We've moved the indexing part in libzim itself and so it left only the mimetype counter.
If we have only one handler, we can probably drop this IHandler and use a more static approach.

I will move code changes in the correct commits but I will not change this right now.
I would like to keep this PR on the adaptation to new api only. Improvement of the code will come in another PRs even if I totally agree with your comment.

src/zimdump.cpp
```diff
@@ -516,20 +414,15 @@ int subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
...
-    return app.dumpArticle(article);
+    try {
+        if (args["--idx"]) {
+            return app.dumpEntry(app.getEntry(args["--idx"].asLong()));
+        } else if(args["--url"]) {
+            return app.dumpEntry(app.getEntryByPath(args["--url"].asString()));
+        }
+    } catch(...) {}
+    std::cerr << "Entry not found" << std::endl;
+    return -1;
```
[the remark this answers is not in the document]
mgautierfr: No, we may go there if there is no `--idx` or `--url` argument.

But I will rewrite this correctly. This is pretty easy to do.

src/zimdump.cpp
```diff
@@ -542,25 +435,18 @@ int subcmdList(ZimDumper &app, std::map<std::string, docopt::value> &args)
...
-        }
+        try {
+            if (idx) {
+                return app.listEntry(app.getEntry(args["--idx"].asLong()));
+            } else if(url) {
+                return app.listEntry(app.getEntryByPath(args["--url"].asString()));
+            }
+        } catch(...) {}
+        std::cerr << "Entry not found" << std::endl;
+        return -1;
```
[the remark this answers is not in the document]
mgautierfr: No, we may go there if there is no `--idx` or `--url` argument.

But I will rewrite this correctly. This is pretty easy to do.

mgautierfr:
Last push rewrite a bit the history :
- fix typos in commit message (path->patch)
- move change in right commit. (In fact split the change between the two commits).

It add also new commits :
- To fix your review (commits to be squash before the merge)
- To fix the unit tests (I realize they was not run on my computer...)

veloman-yunkan (approved):

mgautierfr:
Time to merge this.

PR #199, opened 2020-12-08
Fix zimrecreate to move from a old zim to new one.

mgautierfr:
Need https://github.com/openzim/libzim/pull/462

### DIFF
--- a/src/zimrecreate.cpp
+++ b/src/zimrecreate.cpp
@@ -31,6 +31,66 @@
 #include "tools.h"
 #include "version.h"
 
+/**
+ * A PatchItem. This patch html and css content to remove the namespcae from the links.
+ */
+class PatchItem : public zim::writer::Item
+{
+    //article from an existing ZIM file.
+    zim::Item item;
+
+  public:
+    explicit PatchItem(const zim::Item item):
+      item(item)
+    {}
+
+    virtual std::string getPath() const
+    {
+      auto path = item.getPath();
+      if (path.length() > 2 && path[1] == '/') {
+        path = path.substr(2, std::string::npos);
+      }
+      return path;
+    }
+
+    virtual std::string getTitle() const
+    {
+        return item.getTitle();
+    }
+
+    virtual std::string getMimeType() const
+    {
+        return item.getMimetype();
+    }
+
+    std::unique_ptr<zim::writer::ContentProvider> getContentProvider() const
+    {
+        auto mimetype = getMimeType();
+        if ( mimetype.find("text/html") == std::string::npos
+          && mimetype.find("text/css") == std::string::npos) {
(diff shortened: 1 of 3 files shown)


src/zimrecreate.cpp
```diff
@@ -31,6 +31,59 @@
...
+
+        std::string content = item.getData();
+        // This is a really poor url rewriting to remove the starting "../<NS>/"
+        // and replace the "../../<NS/" by "../"
+        for (std::string prefix: {"'", "\""}) {
+          for (auto ns : {'A','I','J','-'}) {
+            replaceStringInPlace(content, prefix+"../../"+ns+"/", prefix+"../");
+            replaceStringInPlace(content, prefix+"../"+ns+"/", prefix);
+          }
+        }
```
veloman-yunkan, line 82: What does *poor* here imply? If you mean that it can inadvertently modify content that looks like mark-up URL, you better make that explicit.

mgautierfr: I've add a better comment. (And directly squash/rebase as it is the only change)

src/zimrecreate.cpp
```diff
@@ -31,6 +31,66 @@
...
+        // - We may change content starting by `'../A/` even if they are not links
+        // - We don't handle links where we go upper in the middle of the link : `../foo/../I/image.png`
+        // - ...
+        // However, this should patch most of the links in our zim files.
+        for (std::string prefix: {"'", "\""}) {
+          for (auto ns : {'A','I','J','-'}) {
+            replaceStringInPlace(content, prefix+"../../"+ns+"/", prefix+"../");
+            replaceStringInPlace(content, prefix+"../"+ns+"/", prefix);
+          }
+        }
```
veloman-yunkan, line 89: One more case where this may go wrong: `../../<NS>/[AIJ-]/restofurl` will be replaced by `restofurl`.

mgautierfr: Yes, that's why it is a poor algorithm. And there are probably other issue with this.
If we want to do this well, we have to go as in zimchecks : get all links, resolve them to be absolute links, rewrite them as relative links removing the namespaces. (And to be totally sure, we should parse the html, not just get links by looking for `http='<link>'`)

On the other side, zimrecreate is not wildly use (if not use at all), I've create it because I needed to test zim creation without using mwoffliner. No user asked for it and I don't know anybody using it (even in our team).
And situations where the rewrite fails should be very few.

I think we can go with this algorithm for now.
But @kelson42 may have its word to say here.

mgautierfr:
We agree with @kelson42 that the rewrite algorithm was good enough for now. `zimrecreate` is only use by us for testing zim creation, we don't use it in production.

PR #206, opened 2021-01-04
Support for JSON output in zimcheck

### LINKED ISSUE #203: Zimcheck should be able to printout a machine-readable JSON
We are currently deeply integrety the `zimcheck` in the Zimfarm and we the (still to be built) CMS.

If the ZIM check process is part of the Zimfarm process, the level of tolerance regarding errors/warnings is a duty of the CMS.

Therefore the `zimcheck` result will follow that path:
* Zimfarm worker runs zimcheck
* Zimcheck generates a json with all the results (levels, messages, exit code, ...)
* Zimfarm worker upload the json to the zimfarm
* CMS request the zimcheck results to the zimfarm
* CMS decides based on it what to do (publish, alert, throw away the ZIM, etc...)

This is why we need an additional option `--json_log=PATH` which in addition to what is written to the STDOUT/STDIN will write in json PATH the zimcheck output.

veloman-yunkan:
Fixes #203

This is draft/dirty version of the PR. I need feedback before the PR can be finalized.

Producing JSON output while maximally preserving the support for the current format encounters some difficulties because of the heterogeneity of how errors are currently reported:
- The favicon check doesn't issue any message. The message appearing in the report is in a sense the title of a failed check with no other check-specific message following it.
- Some checks (main page, checksum) issue only a single message in case of failure.
- Metadata check can produce a variable though limited number of messages, which could be merged into a single message of variable content.
- Article checks can potentially produce a message for every article. Some of them are "fixed-size" messages (e.g. number of empty links found in an article), while others can be "variable-size" messages (the only example of this is the reporting of dangling links found in an article - the full list of such links is sent via a single message).

Apart from these, the ZIM archive low-level integrity check reports any findings to stderr directly (zimcheck's `ErrorLogger` class 
is not used). This PR doesn't address that check in any way.

The essence of the approach behind this PR is to introduce a table of message templates so that the code for reporting an error only specifies the message id and its parameters. In JSON output mode those parameters are reported as separate properties of the error. In regular mode, the message template is expanded and printed. I used mustache as the template engine. I hit one of its limitations when having to deal with the dangling links error, so in this version of the PR you will see the workaround of passing a list of links as one multi-line string.


mgautierfr:
Shouldn't we use a json library instead of writing it ourselves ?

veloman-yunkan:
I will consider that once the JSON output becomes less trivial.

kelson42:
@veloman-yunkan CI seems broken here?! Anything you should fix?

veloman-yunkan:
> @veloman-yunkan CI seems broken here?! Anything you should fix?

It must be fixed in `kiwix-build` (`mustache` is a new dependency for `zim-tools`)

mgautierfr:
> > @veloman-yunkan CI seems broken here?! Anything you should fix?
> It must be fixed in kiwix-build (mustache is a new dependency for zim-tools)

While I agree it should be fixed in kiwix-build (with https://github.com/kiwix/kiwix-build/pull/470), there is another problem.
http://tmp.kiwix.org/ci/deps2_xenial_native_dyn_zim-tools.tar.xz already contains `mustache.hpp` include file (as the archive contains all base dependencies), so meson should find it.

kelson42:
@veloman-yunkan Any news on this PR?

veloman-yunkan:
@kelson42 The ball is not in my court. I haven't received any feedback to my review request.

kelson42:
@veloman-yunkan OK, I let you deal with @mgautierfr then. I was under the impression, considering that mustache is now properly part of kiwix-build for zim-tools, that you would secure the CI pass.

veloman-yunkan:
> @veloman-yunkan OK, I let you deal with @mgautierfr then. I was under the impression, considering that mustache is now properly part of kiwix-build for zim-tools, that you would secure the CI pass.

@kelson42 @mgautierfr The description of the PR states that this is a draft version and I need some feedback before finalizing it (therefore failing CI shouldn't be a showstopper). I guess my mistake was that I didn't put that as a separate comment when requesting a review. Anyway, I made the Linux builds pass in the CI.

src/zimcheck/checks.cpp
```diff
@@ -127,6 +166,17 @@ void ErrorLogger::report(bool error_details) const {
                 }
             }
         }
+    } else {
```
mgautierfr, line 184: I would prefer a positive test here 
```cpp
if (jsonOutputMode) {
  // Json output
} else {
  // std ouput
}
```

src/zimcheck/checks.cpp
```diff
@@ -296,10 +346,9 @@ void test_articles(const zim::Archive& archive, ErrorLogger& reporter, ProgressB
                     int index = item.getIndex();
                     if (previousIndex != index)
                     {
-                        typedef kainjow::mustache::data data;
-                        data links(data::type::list);
+                        std::string links;
                         for (const auto &olink : p.second)
-                            links << data{"link", olink};
+                            links += "- " + olink + "\n";
```
mgautierfr, line 366: Should we escape the `olink` here ?

test/zimcheck-test.cpp
```diff
@@ -662,8 +664,104 @@ TEST(zimcheck, json_bad_checksum)
...
+      "      'message' : 'article1.html and redundant_article.html',"       "\n"
+      "      'path2' : 'redundant_article.html',"                           "\n"
+      "      'path1' : 'article1.html'"                                     "\n"
+      "    },"                                                              "\n"
+      "    {"                                                               "\n"
+      "      'check' : 'url_internal',"                                     "\n"
+      "      'level' : 'ERROR',"                                            "\n"
+      "      'code' : 6,"                                                   "\n"
+      "      'message' : 'The following links:\\n- A/non_existent.html\\n(/A/non_existent.html) were not found in article dangling_link.html'," "\n"
+      "      'links' : '- A/non_existent.html\\n',"                         "\n"
```
mgautierfr, line 744: I'm not sure to understand the commit message telling we cannot iterate on the links in mustache.

But we should have a json array of `links` here instead of a escaped multi-line string.

mgautierfr (changes requested):

test/zimcheck-test.cpp
veloman-yunkan: The commit message reflects my intention of the same. I added two commit demonstrating the problem. The desired result is achieved with a hack.

mgautierfr: It seems we can iterate over a list : https://mustache.github.io/mustache.5.html (section "Non-Empty Lists")
We need a list of object instead of a list of string but we can. And it is what you've use in your two last commits.

> I will consider that once the JSON output becomes less trivial.

Maybe it is time to consider this not trivial ?

mgautierfr: What do you think of https://github.com/nlohmann/json ?

> With the current approach JSON can be output in a streaming fashion (as each error is generated).

Do we really need that ?
We already wait the end of the process to know all the errors and then we report them.
And if we really need it, we can create a json object per error and stream one object after the other.

src/zimcheck/checks.cpp
veloman-yunkan: This has been obsoleted by the fix of the [other comment](https://github.com/openzim/zim-tools/pull/206#discussion_r589345568). Still, in the modified code no escaping of `olink` is performed since strings are escaped during the JSON output stage.

test/zimcheck-test.cpp
veloman-yunkan: > What do you think of https://github.com/nlohmann/json ?

It looks cute.

> 
> > With the current approach JSON can be output in a streaming fashion (as each error is generated).
> 
> Do we really need that ?
> We already wait the end of the process to know all the errors and then we report them.

If we expect zimcheck to run to an end on a huge ZIM file potentially producing a lot of errors (caused by a buggy scraper or a new check in zimcheck), then we better not accumulate all the errors in memory before printing them. Alternatively we can have a limit on each type of errors.

> And if we really need it, we can create a json object per error and stream one object after the other.

Then we will have to output the enclosing JSON object in a custom way like in the current approach.

---

I don't think that utilizing a json library now will pay off in the short term, but I am fine with doing that and then replacing it with even a better one every time such is found.

veloman-yunkan:
Actually I was more interested in feedback/suggestions related to the difficulties highlighted in the description of the PR:

1. What should be done about the integrity check errors (they are reported to `stderr` directly thus bypassing `ErrorLogger` and hence not showing up in JSON)?
2. What about changing the format of plain-text output so that it is easier to output in JSON mode?  

Also

3. Are numeric error codes useful at all? They are definitely not convenient, especially if their values must be fixed and not change in the future.

test/zimcheck-test.cpp
mgautierfr: > Then we will have to output the enclosing JSON object in a custom way like in the current approach.

Yes, but the home implemented json would simply be a opening/close of a array/dict.
All the complex part (complex object/list, value escaping, ...) would be made by the well known and tested library.

rgaudin:
> > Are numeric error codes useful at all? They are definitely not convenient, especially if their values must be fixed and not change in the future.
> 
> @rgaudin ?

Well @kelson42 might have a different strategy in mind but to me, as long as we can segregate errors and warnings and that we have stable `MsgId`s, we are good to go. zimcheck produces *raw* feedback on specific tests and it's up to the higher-level, more flexible tool to decide what to ultimately do with ZIM files, given this feedback.

Having numeric values would make sense only to sort check results or integrate some hierarchy but it would be painful to implement here and would have good chances of being superseded by that other tool I mentioned.

@kelson42 is that ok with you?

mgautierfr:
@veloman-yunkan What's about this PR ?
Do you plan to use a json library or you don't want to ?

veloman-yunkan:
> @veloman-yunkan What's about this PR ?
> Do you plan to use a json library or you don't want to ?

I would like to avoid dealing with introducing a new dependency in this PR.

veloman-yunkan:
Rebased on master resolving conflicts with #218

veloman-yunkan:
> > @veloman-yunkan What's about this PR ?
> > Do you plan to use a json library or you don't want to ?
> 
> I would like to avoid dealing with introducing a new dependency in this PR.

@mgautierfr What do you think of my approach to dealing with streaming JSON output?

mgautierfr:
> @mgautierfr What do you think of my approach to dealing with streaming JSON output?

Sorry for the delay, I was in holidays.

I would prefer to have a standard and validated library to dump the json.
It would ensure us to have a correct json dumped and avoid use some workaround.
I'm a bit afraid the next time we will change this we will be disturb by this custom jsom dumping.

Adding an external library is not a problem (if we go with the nlohmann/json which is hpp only as mustache).
And we could also use it in kiwix-lib for the suggestions instead of mustache.

But it could be done in another PR. The code is working for now.
Let us concentrate on the functional points (numeric error codes, ...). We will tackle the technical point on json dumping in another PR introducing a json library.

veloman-yunkan:
> Let us concentrate on the functional points (numeric error codes, ...). We will tackle the technical point on json dumping in another PR introducing a json library.

@mgautierfr @kelson42 So how do we proceed from here?

kelson42:
Sorry for the late feedback. I think we need numerical (or at least an other stable string id) error code.... i don’t believe in filtering based on error message.

veloman-yunkan:
@kelson42 BTW, why we need an error code in the first place?

kelson42:
@veloman-yunkan because I want to have each type of error clearly machine readable. This is needed on the CMS side to make a configurable quality filter.

veloman-yunkan:
> @veloman-yunkan because I want to have each type of error clearly machine readable. This is needed on the CMS side to make a configurable quality filter.

I think we can merge this PR without error codes and add them once CMS is ready and the requirements for filtering are better understood.

kelson42:
@veloman-yunkan To me the requirement is clear. It has been stated by me in the ticket clearly 5 months ago, before this PR was even started, see https://github.com/openzim/zim-tools/issues/203#issuecomment-753000670. But this is true that we might need this feature only in a few months as we are still in recrutement process for the CMS developer. If you feel more confortable to split this work, no problem on my side, but then please open a ticket for the part left to do.

mgautierfr:
@veloman-yunkan, I've fixed the mustache detection/correct inclusion but there is a compilation error on MacOs
It seems to no be related to inclusion of mustache but about the json writing.

veloman-yunkan:
> @veloman-yunkan, I've fixed the mustache detection/correct inclusion but there is a compilation error on MacOs
> It seems to no be related to inclusion of mustache but about the json writing.

@mgautierfr I fixed the MacOS build and also dropped the error codes from JSON output.

mgautierfr (approved):

mgautierfr:
> I fixed the MacOS build and also dropped the error codes from JSON output.

So we are good :)

PR #209, opened 2021-01-05
Better openzim/zim-tools image

### LINKED ISSUE #202: Create openzim/zim-tools Docker image
We have currently two Docker images:
* `openzim/zimwriterfs`
* `openzim/checker`

... and theoritically we shoudl create new ones for the other Zim-Tools.

This approach is relatively inefficient because we could have only one `openzim/zim-tools` Docker image which would be easier to maintain.

We want as well to use this opportunity to implement an other approach to build this new Docker image which means:
* Trigger the build based on releases done in `kiwix/build`
* Build with Github Actions and push to Hub+GH Docker image registries
* Use official binaries build (instead of rebuilding within the Dockerfile)
* Use a Github Action parameter to specify the version to build and make the Dockerfile depends on it

Doing all of this will have following benefits:
* Only one Docker image (in place of one per tool)
* Available on both Docker+GH registreries
* Really slim Docker image
* Faster Docker image build process
* Trivial Docker file, no maintenance on the compilation part
* Exactly the same binary online as in the Docker image
* Rebuild automatically for new "builds" as well (not only for new releases)
* Faster Docker image build process

Following that approach we will be able as well to fix https://github.com/kiwix/kiwix-tools/issues/373

rgaudin:
We used to offer a single `openzim/zimwriterfs` image which:

- is named after zimwriterfs alone (although it contains all tools)
- zimwriterfs is sort-of deprecated now
- is a very large image based on Ubuntu and includes compilation tools
- compiles everything from the repo

With this new workflow that releases will trigger we'll get:

- an `openzim/zim-tools` image tagged with the release's tag
- on both dockerhub and ghcr
- using the official binary for that tag

Release can be triggered manually (using GH UI) for any tag, or we can imagine
that kiwix-build triggers it via the API in `build_release_nightly.py`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,26 @@
+name: Docker
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-push:
+    name: Deploy Docker Image
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build and push
+        uses: openzim/docker-publish-action@v3
+        with:
+          image-path: openzim/zim-tools
+          tag-pattern: /^([0-9.]+)$/
+          latest-on-tag: true
+          restrict-to: openzim/zim-tools
+          dockerfile: docker/Dockerfile
+          build-args:
+            VERSION: {version}
+          hub-username: ${{ secrets.DOCKERHUB_USERNAME }}
+          hub-password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          ghcr-username: ${{ secrets.GHCR_USERNAME }}
+          ghcr-token: ${{ secrets.GHCR_TOKEN }}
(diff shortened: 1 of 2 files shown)


kelson42:
>     * zimwriterfs is sort-of deprecated now

Hmmm, we maintain it properly. It is just less critical than before.

kelson42:
> Release can be triggered manually (using GH UI) for any tag, or we can imagine
> that kiwix-build triggers it via the API in `build_release_nightly.py`

Yes and even more important on releases, see https://github.com/kiwix/kiwix-build/issues/465

kelson42 (approved):

rgaudin:
> Hmmm, we maintain it properly. It is just less critical than before.

probably not the right word for it. *not-recommended* maybe? It's *not the advised tool to start with*. I don't think we are still using it on any of our scrapers now.

Issue #200, opened 2020-12-11
Unable to dump to fs properly a ZIM file


kelson42:
```
$ zimdump dump --dir=4242 wiki.openzim.org_en_all_2020-12.zim 
Exception: Error creating symlink from A/wiki.openzim.org/w/images/favicon.ico to 4242/-//favicon
$
```

Here is the ZIM file:
[wiki.openzim.org_en_all_2020-12.zim.zip](https://github.com/openzim/zim-tools/files/5678965/wiki.openzim.org_en_all_2020-12.zim.zip)

maneeshpm:
Hey again @kelson42! if this issue is still up for grabs, I would like to work on it. I was able to reproduce the error, seems like a symlink path error. Can you provide some pointers on how to approach this issue if possible?

kelson42:
@maneeshpm Yes feel free. Unfortunaty I can not really help you but make search in the closed tickets and PR, this will probably be helpful. Fell free to ask as well if you have a question about how the zimdump should behave.

maneeshpm:
Thanks @kelson42, that would be awesome. Meanwhile I can try to pinpoint the exact error.

kelson42:
@maneeshpm This search in the issues/PRs is a work for you, I don't plan to do anything.

maneeshpm:
The exception is generating an `errno 2 ENOENT`. This is caused by the following issues:
- On printing out the path components, for the favicon.ico
`filename: /favicon, directory: 4242, separator: /, retative_path: -//favicon` 
the file name contains a preceding '/', fixed by making `filename = path.substr(position+1);` on line 312. Upon doing this, a valid file path `4242/-/favicon` can be generated.
- Stuck again at the '-' of file path. The `createdir()` function ignores it since `path.size()<=1` and returns. So the directory '-' is never created and hence the ENOENT error.

I guess this might be the issue. After manually creating the '-' directory, things were somewhat different
```
$  zimdump dump --dir=4242 wiki.openzim.org_en_all_2020-12.zim.zip \
Wrote 4242/A/404.html to 4242/_exceptions/A%2f404.html
Wrote 4242/A/index.html to 4242/_exceptions/A%2findex.html
Wrote 4242/A/load.js to 4242/_exceptions/A%2fload.js
Wrote 4242/A/sw.js to 4242/_exceptions/A%2fsw.js
Wrote 4242/A/topFrame.html to 4242/_exceptions/A%2ftopFrame.html
[...]

PR #211, opened 2021-01-07
Fixes issue #200 Unable to dump to fs properly a ZIM file

maneeshpm:
- Fixed redundant separator addition to path.
- Removed `(path.length()<=1)` check in `createdir()` that skipped single character directories.
- Fixed position in substr() inside `createdir()` and `dumpFiles()`.

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -417,12 +417,11 @@ int subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
     std::string directory = args["--dir"].asString();
 
     if (directory.empty()) {
-              throw std::runtime_error(
-                std::string("Directory cannot be empty."));
+        throw std::runtime_error("Directory cannot be empty.");
     }
 
     if (directory.back() == '/'){
-      directory.pop_back();
+        directory.pop_back();
     }
 
     return subcmdDumpAll(app, directory, redirect, filter);
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -416,7 +416,12 @@ int subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
     
     std::string directory = args["--dir"].asString();
 
-    if (*directory.rbegin() == '/'){
+    if (directory.empty()) {
+              throw std::runtime_error(
+                std::string("Directory cannot be empty."));
+    }
+
+    if (directory.back() == '/'){
       directory.pop_back();
     }
 
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -53,9 +53,6 @@ std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
 
 inline static void createdir(const std::string &path, const std::string &base)
 {
-    if (path.size() <= 1)
-        return ;
-
     std::size_t position = 0;
     while(position != std::string::npos) {
         position = path.find('/', position+1);
@@ -347,7 +344,7 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::f
 #else
             if (symlink(redirectPath.c_str(), full_path.c_str()) != 0) {
               throw std::runtime_error(
-                std::string("Error creating symlink from ") + redirectPath + " to " + full_path);
+                std::string("Error creating symlink from ") + full_path + " to " + redirectPath);
             }
 #endif
         }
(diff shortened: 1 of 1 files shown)


maneeshpm:
@kelson42 can you review my pr?

kelson42:
@maneeshpm So directories should never have trailing slashes in zimdump?

kelson42 (changes requested):
A few things need to be improved, clarified.

maneeshpm:
@kelson42
- Adding a check for handling trailing slashes.
- Wrong symlink error message fix. 
- Removed `(path.length()<=1)` from `createdir()`.
An unhandled case is when `--dir=` is left empty, It will generate a symlink error. Are we expecting such a case where a dir is not provided?

src/zimdump.cpp
```diff
@@ -416,7 +413,14 @@ int subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
         std::string nspace = args["--ns"].asString();
         filter = [nspace](const char c){ return nspace.at(0) == c; };
     }
-    return subcmdDumpAll(app, args["--dir"].asString(), redirect, filter);
+    
+    std::string directory = args["--dir"].asString();
+
+    if (*directory.rbegin() == '/'){
```
veloman-yunkan, line 419: 1. What if `directory` is an empty string? (though this is a case for ultimate foolproofness)
2.  Assuming that pre-C++11 compilers are left behind, you can use [`std::string::back()`](https://en.cppreference.com/w/cpp/string/basic_string/back) here, which is more readable.

src/zimdump.cpp
```diff
@@ -325,7 +322,7 @@ void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::f
     }
 
     std::stringstream ss;
-    ss << dir << SEPARATOR << filename;
+    ss << dir << filename;
```
veloman-yunkan, line 325: Now you can simply output `path` here, which means that the need for `ss` above and `relative_path` below has disappeared (or, maybe, `relative_path` can be preserved but simply initialized with `path`, since the name `relative_path` is more appropriate in that code).

Note, that `SEPARATOR` is different between Windows and POSIX, so eliminating it here (and relying on the '/' present in the original path) may not be that insignificant. However, this code contains other possibilities for the POSIX-style path separators to be passed into Windows API calls, so that issue, must be addressed, separately.

veloman-yunkan (changes requested):
This is almost an approval - just a couple of suggestions for slight improvements.

src/zimdump.cpp
maneeshpm: I agree with you @veloman-yunkan, using `back()` is much more readable. If input with an empty string is given, we can either throw a warning `--dir must be provided` or we can dump in the current directory. Which of these is better?

src/zimdump.cpp
maneeshpm: `filename` is a shortened version of the actual filename in `path`. Using `path` directly can cause `createdir()` to fail if the name is too long. eg, this failed:
```
-> dir: A/wiki.openzim.org/w/, 
-> filename: load.phpdebug=false&lang=en&modules=ext.eventLogging.subscriber%7Cext.visualEditor.desktopArticleTarget.init%7Cext.visualEditor.supportCheck%2CtargetLoader%2CtempWikitextEditorWidget%2Ctrack%2Cve%7Cjquery.accessKeyLabel%2CcheckboxShiftClick%2Cclient%2C~1, 
-> path: A/wiki.openzim.org/w/load.php?debug=false&lang=en&modules=ext.eventLogging.subscriber%7Cext.visualEditor.desktopArticleTarget.init%7Cext.visualEditor.supportCheck%2CtargetLoader%2CtempWikitextEditorWidget%2Ctrack%2Cve%7Cjquery.accessKeyLabel%2CcheckboxShiftClick%2Cclient%2Ccookie%2CgetAttrs%2ChighlightText%2Cmw-jump%2Csuggestions%2CtabIndex%2Ctablesorter%2CtextSelection%2Cthrottle-debounce%7Cmediawiki.RegExp%2CString%2CTitle%2CUri%2Capi%2Ccldr%2Clanguage%2Cnotify%2CsearchSuggest%2Cstorage%2Ctemplate%2Cuser%2Cutil%7Cmediawiki.api.options%2Cuser%7Cmediawiki.editfont.styles%7Cmediawiki.language.data%2Cinit%2Cmonths%7Cmediawiki.libs.pluralruleparser%7Cmediawiki.page.ready%2Cstartup%7Cmediawiki.template.regexp%7Cskins.vector.js%7Cuser.defaults&skin=vector&version=1fq761s
```

Yes, `SEPARATOR` should not be removed. Thank you @veloman-yunkan for pointing it out.

src/zimdump.cpp
veloman-yunkan: It is better to reject empty value for `--dir`

src/zimdump.cpp
```diff
@@ -416,7 +416,12 @@ int subcmdDump(ZimDumper &app,  std::map<std::string, docopt::value> &args)
     
     std::string directory = args["--dir"].asString();
 
-    if (*directory.rbegin() == '/'){
+    if (directory.empty()) {
+              throw std::runtime_error(
+                std::string("Directory cannot be empty."));
+    }
```
veloman-yunkan, line 422: 1. Too much indentation
2. Write the throw statement on one line, and omit `std::string` (it is just extra noise when reading the code).

veloman-yunkan (changes requested):

veloman-yunkan (approved):

kelson42 (approved):

PR #212, opened 2021-01-12
Correctly set the Hints for zimrecreate.

mgautierfr:
Needed for (and need) openzim/libzim#466

### DIFF
--- a/src/tools.h
+++ b/src/tools.h
@@ -126,6 +126,10 @@ class ItemProvider : public zim::writer::ContentProvider
     }
 };
 
+// Guess if the item is a front article.
+// This is not a exact science, we use the mimetype to infer it.
+bool guess_is_front_article(const std::string& mimetype);
+
 
 class CopyItem : public zim::writer::Item         //Article class that will be passed to the zimwriter. Contains a zim::Article class, so it is easier to add a
 {
@@ -155,6 +159,10 @@ class CopyItem : public zim::writer::Item         //Article class that will be p
     std::unique_ptr<zim::writer::ContentProvider> getContentProvider() const
     {
        return std::unique_ptr<zim::writer::ContentProvider>(new ItemProvider(item));
+    }
+
+    zim::writer::Hints getHints() const {
+      return { { zim::writer::HintKeys::FRONT_ARTICLE, guess_is_front_article(item.getMimetype()) } };
     }
 };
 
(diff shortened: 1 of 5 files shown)


src/tools.h
```diff
@@ -156,6 +156,10 @@ class CopyItem : public zim::writer::Item         //Article class that will be p
     {
        return std::unique_ptr<zim::writer::ContentProvider>(new ItemProvider(item));
     }
+
+    zim::writer::Hints getHints() const {
+      return {{ zim::writer::HintKeys::FRONT_ARTICLE, (item.getMimetype() == "text/html")}};
+    }
```
kelson42, line 162: @mgautierfr What about if the mimetype has additional attributes like `charset=utf8`? This won't be detected properly?

mgautierfr: I've missed that.
But in the same time, we probably don't want to put `text/html;raw=true` as front article.

Last push fix this.

---

On a more global way, as commented [here](https://github.com/kiwix/kiwix-lib/pull/403#issuecomment-671277426), we face the problem that we have to parse something we shouldn't. (either for `raw=true` than `text/html`)

Hopefully, the (next to implement) extra data feature will fix this.

src/tools.h
```diff
@@ -156,6 +156,13 @@ class CopyItem : public zim::writer::Item         //Article class that will be p
     {
        return std::unique_ptr<zim::writer::ContentProvider>(new ItemProvider(item));
     }
+
+    zim::writer::Hints getHints() const {
+      auto mimetype = item.getMimetype();
+      auto front_article = ( mimetype.find("text/html") != std::string::npos
+                          && mimetype.find("raw=true") == std::string::npos);
```
veloman-yunkan, line 163: Convert this piece of code to a free helper function `bool is_front_article(const zim::Item& item)` and reuse it in `PatchItem`.

veloman-yunkan (changes requested):

src/tools.h
mgautierfr: Done. Using name `guess_is_front_article` because it is a guess. mimetype is not a real criteria, but we don't have best.

veloman-yunkan (approved):

kelson42:
@mgautierfr I have rebased, but testing seems insufficient

mgautierfr:
I've updated a bit the code :
- The helper function now takes a "mimetype" string instead of a item. It allows us to unit test it.
- The mimetype must start by "text/html", not just containing it.
- I have added unit test on method `guess_is_front_article`

The coverage is pretty low as we don't run zim-recreate at all in the unit tests.
And all the "writerItem" methods are called by `zim::writer::Creator` when we actually recreate a zim.

veloman-yunkan (approved):

PR #215, opened 2021-01-17
Almost full coverage of zimcheck by unit tests

veloman-yunkan:
Before implementing support for JSON output in zimcheck (ticket #203, PR #206) we must ensure almost full coverage of zimcheck code, or else the risk of breaking zimcheck by the said enhancement will be quite high. This PR achieves that goal. While doing so, several bugs were found and fixed too.

A couple of things to note about this PR:

1. It introduces the same kind of issues must be solved by openzim/libzim#469. However the PR has been cleaned in such a way that it introduces only 3 ZIM files with a total additional load on git of around 160KiB (each ZIM file is added only once and is not modified later).
2. `zimwriterfs` has been modified so that it can produce ZIM files suitable for testing `zimcheck`:
    - A hidden option `--dont-check-arguments|-B` was added to `zimwriterfs`. In that mode the latter doesn't validate some of its arguments, and it is possible to generate ZIM files that trigger zimcheck errors. 
    - Next, an attempt was made to produce ZIM files deterministically. To this end options `--no-uuid` and `--threads` were added to `zimwriterfs` (`--no-uuid` is hidden, while `--threads` is public). The main purpose was to get rid of the randomness in UUID so that it can be put in the expected output of unit tests of JSON mode. A secondary goal was to make sure that running `zimwriterfs` several times on the same input produces the same output, so that the test ZIM file generation procedure doesn't have to include complex logic of which ZIM files should be left untouched and which ones need to be updated. Running `zimwriterfs` single-threaded (via `--threads 1`) brings us closer to that goal, however doesn't fully achieve it. There are at least two known remaining sources of randomness in a ZIM file created by `zimwriterfs`: the UUID generated and stored in its database by Xapian and the creation date written in metadata (the latter can be solved but the former is out of our control, unless patching Xapian is acceptable).
4. The ZIM files in this PR were generated using a build of `zimwriterfs` corresponding to this PR.
5. The test ZIM file creation script [test/data/zimfiles/create_test_zimfiles](https://github.com/openzim/zim-tools/blob/more-zimcheck-unit-tests/test/data/zimfiles/create_test_zimfiles) contains a [hardcoded required version](https://github.com/openzim/zim-tools/blob/more-zimcheck-unit-tests/test/data/zimfiles/create_test_zimfiles#L26) for `zimwriterfs`. The version is incremented by this PR.


kelson42:
@veloman-yunkan Thank you very much for your effort of pushing testing. I know how a hassle it is but how necessary it is. That said, I'm not in favour of having user-visible options which are only necessary for dev purpose. This is not a good pattern. We should have another way to do this tests without overloading the user with options he will never use.

veloman-yunkan:
@kelson42 Is it OK if I just remove the dev options from the usage/help message?

kelson42:
@veloman-yunkan To me on the limit of what is acceptable... but lets go for it :)

veloman-yunkan:
> @veloman-yunkan To me on the edge of what is acceptable... but lets go for it :)

Done. Rebased/rewrote history since I also fixed a small issue in the source data of the test ZIM files and regenerated them while rebasing.

veloman-yunkan:
BTW, I left the `--threads` option of `zimwriterfs` public.

src/zimwriterfs/zimwriterfs.cpp
```diff
@@ -316,6 +323,7 @@ void parse_args(int argc, char** argv)
     usage();
     exit(1);
   }
+  }
```
mgautierfr, line 326: Shouldn't we check the directoryPath and zimPath even if we have `dontCheckArgs` ?

test/zimcheck-test.cpp
```diff
@@ -87,6 +87,27 @@ class CapturedStdout
...
+  CapturedStderr(const CapturedStderr&) = delete;
+
+  ~CapturedStderr()
+  {
+    std::cerr.rdbuf(sbuf);
+  }
+
+  operator std::string() const { return buffer.str(); }
+};
+
```
mgautierfr, line 110: It would be better to make `CapturedStdOut` configurable (and rename it to `CatpturedStdStream`) than creating almost the same class.

mgautierfr (changes requested):

src/zimwriterfs/zimwriterfs.cpp
veloman-yunkan: Fixed. Also noticed and fixed an error in the short option corresponding to `--dont-check-arguments` (I had initially used `-g`, then changed it to `-B` but forgot to update the call to `getopt_long()`).

test/zimcheck-test.cpp
veloman-yunkan: Done, but I preserved `CapturedStdout` and `CapturedStderr` as thin wrappers around `CapturedStdStream` in order to keep intact the 16 locations where they were used. In my opinion,

```
CapturedStdout stdout;
```
is more readable (less noisy) than

```
CapturedStdStream stdout(std::cout);
```

mgautierfr:
We are good with the last changes.

mgautierfr (approved):

veloman-yunkan:
Done.

PR #225, opened 2021-02-19
Correct relative path in HTTP redirect generated by zimdump

### LINKED ISSUE #224: zimdump produces invalid path in HTML-based redirect
> - This is something I've found while working on https://github.com/ipfs/distributed-wikipedia-mirror/issues/60 
> - Not a hard blocker, will fix this for names without `/` in additional build step in https://github.com/ipfs/distributed-wikipedia-mirror/pull/77, but nice to fix it upstream in a way that works for all article names.

Assuming I understood this correctly, the HTML-based redirect created by `zimdump`  points at an invalid path.

### How to reproduce

`wikipedia_cr_all_maxi_2021-02.zim` is small enough to be a good demo:

```console
$ zimdump dump --dir=cr-tmp ./wikipedia_cr_all_maxi_2021-02.zim
$ cat cr-tmp/A/English_language
<!DOCTYPE html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="0;url=A/%E1%90%A7%E1%90%81%E1%92%A5%E1%94%A5%E1%91%8E%E1%91%AF%E1%94%92%E1%90%A4_%E1%90%8A%E1%94%A8%E1%92%A7%E1%90%A7%E1%90%83%E1%93%90" /><head><body></body></html>
```

So while in `A/{name1}` we are redirected to `A/{name2}`.

This looks like a bug, because the redirect is relative to `{name1}`, so it points at `A/A/{name2}` which does not exist.  

I was able to reproduce the same issue with 
- `wikipedia_cu_all_maxi_2021-02` – example: `A/Жена` effectively redirects to  `A/A/%D0%96%D1%94%D0%BD%D0%B0`
- `wikipedia_tr_all_maxi_2021-02`  – example:  `A/Fatih_Sultan_Mehmed` effectively redirects to `A/A/II._Mehmed`

#### Relative redirect

Ideally, a relative path would be used, without namespace prefix,  something like:

```diff
- <meta http-equiv="refresh" content="0;url=A/{name2}" /> 
+ <meta http-equiv="refresh" content="0;url={name2}" /> 
```

#### Open problem: subdirectories

The caveat is an edge case when a name with `/` is redirecting to something else.  Something like  `A/some/name`  needs to point at article one level up:

```diff
- <meta http-equiv="refresh" content="0;url=A/{name2}" /> 
+ <meta http-equiv="refresh" content="0;url=../{name2}" /> 
```

In case that  adds too much complexity to `zimdump` code,  we could  do that in JS.

If we replace `<meta>` tag with `<script>` then we are able to correctly redirect, no matter if name includes `/` or not. Something like: 

```html
<!doctype html>
<html lang="en">
<head>
<script>
  const from = 'na/me1'
  const to = 'name2'
  const prefix = window.location.pathname.split(from)[0]  
  window.location.replace(prefix + to)
</script>
<noscript>
  <meta http-equiv="refresh" content="0;url=name2" />  <!-- best-effort if no JS -->
</noscript>
</head>
</html>
```

@kelson42 lmk if there is a better way (or if I misunderstood how things work :))

veloman-yunkan:
Fixes #224

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -25,6 +25,7 @@
 #include <sys/stat.h>
 #include <cerrno>
 #include <fstream>
+#include <iomanip>
 #include <iostream>
 #include <sstream>
 #include <stdexcept>
@@ -33,6 +34,7 @@
 #include <unistd.h>
 #include <algorithm>
 #include <regex>
+#include <array>
 
 #ifdef _WIN32
 #define SEPARATOR "\\"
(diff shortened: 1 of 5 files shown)


src/tools.cpp
```diff
@@ -153,6 +153,50 @@ static std::string removeLastPathElement(const std::string& path,
...
+      r += "../";
+    }
+
+    for ( auto it = x.second; it != b.end(); ++it ) {
+        if ( it != x.second )
+          r += "/";
+        r += *it;
+    }
+    return r;
+}
```
mgautierfr, line 198: kiwix-lib already have a `computeRelativePath` (https://github.com/kiwix/kiwix-lib/blob/master/src/tools/pathTools.cpp#L176-L199).
The kiwix one does a bit more as it normalize the path (remove unnecessary internal `..`).
But this implementation is a bit simpler (use of `mismatch` instead of reimplement it)
It would be nice to have some synchronization (or even move the `computeRelativePath` in libzim and use only this one)

I'm not sure this implementation can handle windows path (but we are looking for `/` in the path elsewhere in the code, to the questioning is more global to all zimdump tool)

src/zimdump.cpp
```diff
@@ -302,6 +306,17 @@ inline void write_to_file(const std::string &base, const std::string& path, cons
...
+{
+    redirectPath = computeRelativePath(currentEntryPath, redirectPath);
+    auto encodedurl = urlEncode(redirectPath, true);
+    std::ostringstream ss;
+
+    ss << "<!DOCTYPE html><head><meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\" />";
+    ss << "<meta http-equiv=\"refresh\" content=\"0;url=" + encodedurl + "\" /><head><body></body></html>";
+    auto content = ss.str();
+    write_to_file(directory + SEPARATOR, outputPath, content.c_str(), content.size());
+}
```
mgautierfr, line 319: It would be nice to test this, even if you have not written it (to reach the 80% coverage).
Maybe move the generation of the content in a free function to simplify the test.

mgautierfr (changes requested):

src/zimdump.cpp
veloman-yunkan: That was a good advice. The HTML was buggy (not opening `<html>` tag, `<head>` was not properly closed).

src/tools.cpp
veloman-yunkan: I don't think anything needs to be done about it in this PR.

mgautierfr (approved):

PR #228, opened 2021-02-27
fix(zimdump): add canonical links to redirects

lidel:
Without this search engines crawling unpacked version may produce duplicated results.
It does not cost much to have it, and avoids issues like ones linked below.

Ref.
https://github.com/openzim/mwoffliner/pull/963
https://github.com/ipfs/distributed-wikipedia-mirror/issues/65
https://en.wikipedia.org/wiki/Canonical_link_element

cc @kelson42

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -507,6 +507,7 @@ std::string httpRedirectHtml(const std::string& redirectUrl)
           "<head>"
           "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\" />"
           "<meta http-equiv=\"refresh\" content=\"0;url=" + encodedurl + "\" />"
+          "<link rel=\"canonical\" href=\"" + encodedurl + "\" />"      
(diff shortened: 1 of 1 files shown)


mgautierfr:
From a mwoffliner, the redirect comes from a "real" redirect in the source wiki and so have a canonical link we know at zim file creation (but lost in the process).
This canonical link will be different that the real one.
(Just want to mention. We have no way to solve this for now)

kelson42:
@lidel I don’t understand the concrete problem this PR should fix. Would you be able please to open the corresponding issue ticket? In general we don’t recommend to create a PR without having a proper issue ticket, then we shoukd reach consense that there is a need to do something before starting inplementing something.

kelson42:
@lidel ?

lidel:
@kelson42 apologies for missing this, you were right to close this (I was mixing concerns, this does not belog to zimdump)

PR #229, opened 2021-03-01
More robust `computeRelativePath()`

### LINKED ISSUE #227: Memory leak in the latest zimdump nightly (2021-02-27)
`zimdump` from https://download.openzim.org/nightly/2021-02-27/  seems to have a memory leak when processing [`wikipedia_my_all_maxi_2021-02.zim`](https://download.kiwix.org/zim/wikipedia/wikipedia_my_all_maxi_2021-02.zim):

```console
$ zimdump dump ./snapshots/wikipedia_my_all_maxi_2021-02.zim --dir ./tmp/wikipedia_my_all_maxi_2021-02
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A//i/ to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2f%2fi%2f
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A//y/ to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2f%2fy%2f
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A//ø/ to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2f%2fø%2f
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A//ɪ/ to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2f%2fɪ%2f
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A//ʏ/ to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2f%2fʏ%2f
Wrote ./tmp/wikipedia_my_all_maxi_2021-02/A/C/2020_F3_(NEOWISE) to ./tmp/wikipedia_my_all_maxi_2021-02/_exceptions/A%2fC%2f2020_F3_(NEOWISE)
zsh: killed     zimdump dump ./snapshots/wikipedia_my_all_maxi_2021-02.zim --dir 
zimdump dump ./snapshots/wikipedia_my_all_maxi_2021-02.zim --dir   43.73s user 5.48s system 99% cpu 49.480 total
```

kernel log:
```
Feb 27 23:16:00 [kernel] [89985.314555] Out of memory: Killed process 13126 (zimdump) total-vm:47336948kB, anon-rss:25586020kB, file-rss:4kB, shmem-rss:0kB, UID:1000 pgtables:50148kB oom_score_adj:0
Feb 27 23:16:00 [kernel] [89985.635024] oom_reaper: reaped process 13126 (zimdump), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

This is is not present in https://download.openzim.org/nightly/2021-02-12/  which I had around and re-tested with, so there is most-likely some regression intorduced by https://github.com/openzim/zim-tools/pull/225 (there were no other changes afaik).

@kelson42 I tested #225 against smaller  `wikipedia_cu_all_maxi_2021-02.zim` – that worked fine, so I  missed this before

veloman-yunkan:
Should fix #227 

I didn't test the fix on the ZIM file provided in the linked issue. Instead the problem was identified via code review. When the two paths in `computeRelativePath()` are the same (or the first one is the path of the directory containing the second one), the unsigned variable `ups` is assigned a value of -1 (which, depending on the platform, underflows to 2³²-1 or 2⁶⁴-1):

https://github.com/openzim/zim-tools/blob/f406219cd974d2a944cccbc72a0da8616d886972/src/tools.cpp#L179-L192

The issue seems obvious enough to avoid downloading a 800MiB file.

Since the inputs to `computeRelativePath()` have a semantics of URLs rather than file system paths, I also added testing (and, therefore, addressed handling) of various edge cases related to the usage of a [trailing slash](https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/).

### DIFF
--- a/test/tools-test.cpp
+++ b/test/tools-test.cpp
@@ -57,17 +57,47 @@ TEST(CommonTools, decodeUrl)
 
 TEST(CommonTools, computeRelativePath)
 {
-  EXPECT_EQ("B", computeRelativePath("A", "B"));
+  EXPECT_EQ("A",  computeRelativePath("A" , "A" ));
+  EXPECT_EQ("..", computeRelativePath("A/", "A" ));
+  EXPECT_EQ("A/", computeRelativePath("A" , "A/"));
+  EXPECT_EQ("./", computeRelativePath("A/", "A/"));
+
+  EXPECT_EQ("B",     computeRelativePath("A" , "B" ));
+  EXPECT_EQ("../B",  computeRelativePath("A/", "B" ));
+  EXPECT_EQ("B/",    computeRelativePath("A" , "B/"));
+  EXPECT_EQ("../B/", computeRelativePath("A/", "B/"));
+
+  EXPECT_EQ("A/B",  computeRelativePath("A" , "A/B" ));
+  EXPECT_EQ("B",    computeRelativePath("A/", "A/B" ));
+  EXPECT_EQ("A/B/", computeRelativePath("A" , "A/B/"));
+  EXPECT_EQ("B/",   computeRelativePath("A/", "A/B/"));
+
+  EXPECT_EQ("..",     computeRelativePath("A/B",  "A" ));
+  EXPECT_EQ("../..",  computeRelativePath("A/B/", "A" ));
+  EXPECT_EQ("../",    computeRelativePath("A/B",  "A/"));
+  EXPECT_EQ("../../", computeRelativePath("A/B/", "A/"));
+
   EXPECT_EQ("B/CD/EFG", computeRelativePath("A", "B/CD/EFG"));
 
   EXPECT_EQ("c", computeRelativePath("dir/b", "dir/c"));
+  EXPECT_EQ("b", computeRelativePath("dir/b", "dir/b"));
+  EXPECT_EQ("b/c", computeRelativePath("dir/b", "dir/b/c"));
+  EXPECT_EQ("..", computeRelativePath("dir/b/c", "dir/b"));
   EXPECT_EQ("subdir/", computeRelativePath("dir/b", "dir/subdir/"));
+  EXPECT_EQ("../", computeRelativePath("dir/subdir/b", "dir/subdir/"));
+  EXPECT_EQ("..", computeRelativePath("dir/subdir/b", "dir/subdir"));
+  EXPECT_EQ("../../", computeRelativePath("dir/subdir/b/c", "dir/subdir/"));
+  EXPECT_EQ("../..", computeRelativePath("dir/subdir/b/c", "dir/subdir"));
   EXPECT_EQ("../c", computeRelativePath("dir/subdir/", "dir/c"));
 
   EXPECT_EQ("../c", computeRelativePath("A/B/f", "A/c"));
   EXPECT_EQ("D/c", computeRelativePath("A/f", "A/D/c"));
 
   EXPECT_EQ("c", computeRelativePath("A/B/f", "A/B/c"));
+  EXPECT_EQ("c", computeRelativePath("A/B/c", "A/B/c"));
+  EXPECT_EQ("c/d", computeRelativePath("A/B/c", "A/B/c/d"));
+  EXPECT_EQ("..", computeRelativePath("A/B/c/d", "A/B/c"));
+  EXPECT_EQ("../..", computeRelativePath("A/B/c/d", "A/B"));
 }
 
 TEST(CommonTools, computeAbsolutePath)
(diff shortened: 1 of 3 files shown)


kelson42:
@mgautierfr Please review this PR today as this is stringly impairing for @lidel and the Wikipedia IPFS project.

test/tools-test.cpp
```diff
@@ -57,17 +57,47 @@ TEST(CommonTools, decodeUrl)
...
+  EXPECT_EQ("A/B",  computeRelativePath("A" , "A/B" ));
+  EXPECT_EQ("B",    computeRelativePath("A/", "A/B" ));
+  EXPECT_EQ("A/B/", computeRelativePath("A" , "A/B/"));
+  EXPECT_EQ("B/",   computeRelativePath("A/", "A/B/"));
+
+  EXPECT_EQ("..",     computeRelativePath("A/B",  "A" ));
+  EXPECT_EQ("../..",  computeRelativePath("A/B/", "A" ));
+  EXPECT_EQ("../",    computeRelativePath("A/B",  "A/"));
+  EXPECT_EQ("../../", computeRelativePath("A/B/", "A/"));
+
```
mgautierfr, line 79: Can you add some tests (the same ?) with leading "/"  ?

mgautierfr (changes requested):
> P.S. Does computeRelativePath() in libkiwix have to deal with that peculiarity related to the trailing slash?

This already deal with it :
- We already split the paths explicitly keeping "empty" part. So we trailing slash will generate a empty part at the end.
- This is tested here : https://github.com/kiwix/kiwix-lib/blob/master/test/pathTools.cpp#L145-L146, https://github.com/kiwix/kiwix-lib/blob/master/test/pathTools.cpp#L149-L150 (although tests may not be exhaustive)

test/tools-test.cpp
veloman-yunkan: Currently `computeRelativePath()` is used only with paths returned by `zim::Entry::getPath()` which don't contain a leading slash.

veloman-yunkan:
> > P.S. Does computeRelativePath() in libkiwix have to deal with that peculiarity related to the trailing slash?
> 
> This already deal with it :
> 
>  * We already split the paths explicitly keeping "empty" part. So we trailing slash will generate a empty part at the end.
> 
>  * This is tested here : https://github.com/kiwix/kiwix-lib/blob/master/test/pathTools.cpp#L145-L146, https://github.com/kiwix/kiwix-lib/blob/master/test/pathTools.cpp#L149-L150 (although tests may not be exhaustive)

Those tests demonstrate that the trailing slash is handled properly but the absence of a trailing slash isn't (as long as URLs are concerned).

[For example](https://github.com/kiwix/kiwix-lib/blob/7ece3830048cbe60924bbc36a22fc39bf4fbf95f/test/pathTools.cpp#L147-L150)

```c++
  ASSERT_EQ(computeRelativePath(A2("a","b"), A2("a","foo")),
            P2("..","foo"));
  ASSERT_EQ(computeRelativePath(A3("a","b",""), A2("a","foo")),
            P2("..","foo"));
```

which is equivalent in Linux/URL style to

``` c++
  ASSERT_EQ(computeRelativePath("a/b",  "a/foo"), "../foo"); // !!! wrong, should be "./foo" or "foo"
  ASSERT_EQ(computeRelativePath("a/b/", "a/foo"), "../foo"); // correct
```

test/tools-test.cpp
veloman-yunkan: The comment above `computeRelativePath()` also states that the input paths are expected to be relative paths:  https://github.com/openzim/zim-tools/blob/ff61a93e2219407f25d31494e722a9b98d35488b/src/tools.h#L165-L168

mgautierfr:
You are right. Although this function is used only once to compute the relative link when generated library xml. It is always used (for now) with the first argument always being a directory.

But anyway, this should be discussed in kiwix-lib repository. It has no impact on this PR.
I'm merging it.

mgautierfr (approved):

mgautierfr:
I've created kiwix/kiwix-lib#457 to track the issue with kiwix-lib's computeRelativePath.

lidel:
This will land in next [nightly](https://download.openzim.org/nightly/) for 2021-03-03, right?

kelson42:
@lidel Yes it should, to be sure check the file generation date against the merge date.

lidel:
Thanks, I tested nightly for 2021-03-03 and confirmed the issue is fixed – was able to unpack  `wikipedia_my_all_maxi_2021-02` without any issues.

Issue #242, opened 2021-05-16
Installing zim-tools fails on Ubuntu Server 20.04.2 -- partial workaround is to download mustache.hpp into /usr/include


holta:
Summary: zim-tools install instructions at https://github.com/openzim/zim-tools#compilation do not work on Ubuntu Server 20.04.2

_Thanks if anybody has suggestions for PROBLEM 2 OF 2 below!_

### PROBLEM 1 OF 2:

```
git clone https://github.com/openzim/zim-tools
cd zim-tools
meson . build
```

...can be solved with this workaround: (thanks to @curtathompson and @georgejhunt)

```
cd /usr/include
wget https://raw.githubusercontent.com/kainjow/Mustache/master/mustache.hpp
cd /opt/iiab/zim-tools
meson . build
```

Error Log 1: (output of `meson . build`)
https://paste.ubuntu.com/p/wZkcfSHsTm/

Error Log 2: (/opt/iiab/zim-tools/build/meson-logs/meson-log.txt)
https://paste.ubuntu.com/p/HK45rXKqFS/

### PROBLEM 2 OF 2: (this is the output of `ninja -C build`)

```
ninja: Entering directory `build'
[2/22] Compiling C++ object 'src/25a6634@@zimsearch@exe/zimsearch.cpp.o'.
FAILED: src/25a6634@@zimsearch@exe/zimsearch.cpp.o
c++ -Isrc/25a6634@@zimsearch@exe -Isrc -I../src -I/usr/local/include -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Werror -std=c++11 -g -Werror -Wall '-DVERSION="2.2.0"' -MD -MQ 'src/25a6634@@zimsearch@exe/zimsearch.cpp.o' -MF 'src/25a6634@@zimsearch@exe/zimsearch.cpp.o.d' -o 'src/25a6634@@zimsearch@exe/zimsearch.cpp.o' -c ../src/zimsearch.cpp
../src/zimsearch.cpp: In function ‘void printSearchResults(zim::Search&)’:
../src/zimsearch.cpp:28:23: error: ‘iterator’ is not a member of ‘zim::Search’
   28 |     for (zim::Search::iterator it = search.begin(); it != search.end(); ++it)
      |                       ^~~~~~~~
../src/zimsearch.cpp:28:53: error: ‘it’ was not declared in this scope; did you mean ‘int’?
[...]

PR #243, opened 2021-05-17
Adapt zim-tools to new libzim api.

mgautierfr:
This adapt zim tools to what have been already merged in libzim.

### DIFF
--- a/src/zimwriterfs/zimwriterfs.cpp
+++ b/src/zimwriterfs/zimwriterfs.cpp
@@ -154,7 +154,7 @@ void usage()
   std::cout << "\t-h, --help\t\tprint this help" << std::endl;
   std::cout << "\t-V, --version\t\tprint the version number" << std::endl;
   std::cout
-      << "\t-m, --minChunkSize\tnumber of bytes per ZIM cluster (default: 2048)"
+      << "\t-m, --minChunkSize\tnumber of kilobytes per ZIM cluster (default: 2048)"
       << std::endl;
   std::cout << "\t-J, --threads\tcount of threads to utilize (default: 4)"
       << std::endl;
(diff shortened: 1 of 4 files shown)


src/zimwriterfs/zimwriterfs.cpp
```diff
@@ -154,7 +154,7 @@ void usage()
   std::cout << "\t-h, --help\t\tprint this help" << std::endl;
   std::cout << "\t-V, --version\t\tprint the version number" << std::endl;
   std::cout
-      << "\t-m, --minChunkSize\tnumber of bytes per ZIM cluster (default: 2048)"
+      << "\t-m, --minChunkSize\tnumber of kilobytes per ZIM cluster (default: 2048)"
```
veloman-yunkan, line 157: What about renaming this option?

holta:
I tested this PR.

It works to install zim-tools and zimdump in the context of #242.

src/zimwriterfs/zimwriterfs.cpp
mgautierfr: Done.

kelson42:
@mgautierfr Would be great if you could put a word about Mustache in the README.md as well.

mgautierfr:
> Would be great if you could put a word about Mustache in the README.md as well.

It is two different issues.
Here we are fixing compilation dues to a change in libzim. (so zim-tools has to adapt)
Mustache is because we add the dependency in a PR of zim-tools itself.

veloman-yunkan (changes requested):
`long_options` were not updated. I guess you missed it because `minchunksize` is all lowercase there.

mgautierfr:
> long_options were not updated. I guess you missed it because minchunksize is all lowercase there.

Yes. And I've test with `-clusterSize`, which didn't raise a error as it is interpreted as `creator` option :facepalm: 

It is fixed (and already fixup in the commit).

veloman-yunkan (approved):

PR #253, opened 2021-07-17
Support for new-style illustrations in zimcheck

### LINKED ISSUE #251: [zimcheck] Illustration not detected as favicon
Looks like zimcheck is only looking for old-fashion favicon and not for Illustrations.

Example with [this zim file](http://tmp.kiwix.org/beer_meta_stackexchange_com_2021-07.zim) containing 2 illustrations ; properly shown by latest kiwix-serve.

veloman-yunkan:
Fixes #251

In the current fix, the existence of any new-style illustration results in the favicon check being satisfied. 

Questions:

1. Is it acceptable to trust the `zim::Archive::getIllustrationSizes()` function or `zimcheck` should rather check for the favicons using its own explicit logic?
2. If we decide to trust the `zim::Archive::getIllustrationSizes()` function, should the old code be preserved? Note that `zim::Archive::getIllustrationSizes()` contains the same logic for old-style ZIM archives.
3. Should we check that the size required by the spec is present?

### DIFF
--- a/src/zimcheck/checks.cpp
+++ b/src/zimcheck/checks.cpp
@@ -280,7 +280,8 @@ void test_metadata(const zim::Archive& archive, ErrorLogger& reporter) {
 void test_favicon(const zim::Archive& archive, ErrorLogger& reporter) {
     reporter.infoMsg("[INFO] Searching for Favicon...");
 
-    if ( archive.getIllustrationSizes().empty() )
+    const auto illustrationSizes = archive.getIllustrationSizes();
+    if ( illustrationSizes.find(48) == illustrationSizes.end() )
       reporter.addMsg(MsgId::MISSING_FAVICON, {});
 }
 
--- a/src/zimcheck/checks.cpp
+++ b/src/zimcheck/checks.cpp
@@ -279,13 +279,9 @@ void test_metadata(const zim::Archive& archive, ErrorLogger& reporter) {
 
 void test_favicon(const zim::Archive& archive, ErrorLogger& reporter) {
     reporter.infoMsg("[INFO] Searching for Favicon...");
-    static const char* const favicon_paths[] = {"-/favicon.png", "I/favicon.png", "I/favicon", "-/favicon"};
-    for (auto &path: favicon_paths) {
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin Would you be able please to confirm this PR fix "your" problem?

kelson42:
@veloman-yunkan I will try to anser:
1 - I would trust (at least in a first time)... not really eager to build a dependence to libpng in zimcheck
2 - I don't really understand what is meant here as I don't know what is special with current version.
3 - yes please, we ahoukd have at least 48x48

rgaudin:
Works for me 👍 

```
./build/src/zimcheck/zimcheck -F ~/data/stackexchange/3dprinting_stackexchange_com_2021-07.zim
[INFO] Checking zim file /Users/reg/data/stackexchange/3dprinting_stackexchange_com_2021-07.zim
[INFO] Overall Test Status: Pass
[INFO] Total time taken by zimcheck: 0 seconds.
```

mgautierfr:
> Is it acceptable to trust the zim::Archive::getIllustrationSizes() function or zimcheck should rather check for the favicons using its own explicit logic?

Yes. We already trust libzim to get the metadata or find the indexes. We should also trust it for the illustration (or never).

> If we decide to trust the zim::Archive::getIllustrationSizes() function, should the old code be preserved? Note that zim::Archive::getIllustrationSizes() contains the same logic for old-style ZIM archives.

No, `getIllustrationSizes` does the job.

> Should we check that the size required by the spec is present?

Yes.
`zimcheck` must check that the content in the zim file is correct/coherent.
We must check that we respect the spec:
- At least that we have declared illustration with the correct size, required by the spec,
- Ideally, check that the actual data correspond to what is declared (png, correct size/scale). But as @kelson42 said this will add a dependency to a png library. This may wait another PR.

mgautierfr (commented):
Please remove the old code to get the favicon but else we are good.

mgautierfr (approved):
We are good.
Please rebase-fixup before we merge.

veloman-yunkan:
> Please rebase-fixup before we merge.

Done

PR #275, opened 2022-01-02
Better version

kelson42:
Print dependences library version number as well if -v, --version is call for any command line zim tool.

### DIFF
--- a/src/version.h
+++ b/src/version.h
@@ -28,8 +28,8 @@
   #define VERSION "undefined"
 #endif
 
-void version(std::ostream& out = std::cout) {
-  out << VERSION << std::endl;
+void printVersions(std::ostream& out = std::cout) {
+  out << "zim-tools " << VERSION << std::endl;
   out << std::endl;
   zim::printVersions(out);
 }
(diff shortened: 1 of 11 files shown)


src/version.h
```diff
@@ -28,7 +28,7 @@
   #define VERSION "undefined"
 #endif
 
-void version(std::ostream& out = std::cout) {
+void printVersions(std::ostream& out = std::cout) {
   out << VERSION << std::endl;
```
veloman-yunkan, line 32: Let's start the versions text with "zim-tools"

veloman-yunkan (changes requested):

src/version.h
veloman-yunkan: https://github.com/kiwix/kiwix-tools/pull/516#discussion_r777203400 is an answer to this comment too

veloman-yunkan (approved):

src/version.h
kelson42: Like on kiwix-tools, finally changed my mind.

PR #208, opened 2021-01-05
Bump-up version (and libzim required version)

kelson42:
To avoid #207
Need https://github.com/openzim/libzim/pull/468 to be merged first

PR #283, opened 2022-01-19
Version 3.1.0

mgautierfr:


test/zimcheck-test.cpp
```diff
@@ -254,7 +254,7 @@ TEST(zimcheck, integrity_goodzimfile)
 {
     const std::string expected_output(
         "[INFO] Checking zim file data/zimfiles/good.zim" "\n"
-        "[INFO] Zimcheck version is 3.0.0" "\n"
+        "[INFO] Zimcheck version is " VERSION "\n"
         "[INFO] Verifying ZIM-archive structure integrity..." "\n"
```
veloman-yunkan, line 258: This change means that we don't care about the version string in the zimcheck output (since we test the contents of the VERSION macro via itself). Consider that the `VERSION` macro is carelessly redefined (`#undef`ed and then `#define`d)  by a header file included *after* `zim-tools/src/version.h`. Then our tests will not detect that.  

This comment also applies to an earlier commit 409824a44cc500d73016ac038b3cbf4df1a0f2e4 (in #208) but then we at least had a fallback in the form of this kind of checks.

I think we should preserve the version in an explicit textual form in at least one place (and better in three - output of `zimcheck -v`, version info in plaintext output and version info in JSON output).

mgautierfr: Indeed.
I was more on testing the output itself than the version.
I have created the issue #285 for this.

PR #288, opened 2022-02-05
Small runs of zimcheck are now spreading love

### LINKED ISSUE #287: Tests fail if the machine is slow
Some of the Debian buildds are virtualized hardware and can be slow, so the test fails with:

```
Expected equality of these values:
...
-[INFO] Total time taken by zimcheck: 0 seconds.\n
+[INFO] Total time taken by zimcheck: 1 seconds.\n
```

Can we fix the tests so they still pass regardless of how long it takes?

Filed downstream as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1004900

veloman-yunkan:
Fixes #287

The unit-tests of zimcheck include a report on the total runtime of each flow. Since those flows are very small the corresponding line from the zimcheck output used to be

    [INFO] Total time taken by zimcheck: 0 seconds.

which corresponds to the duration of the flow being under half a second.

However, Debian builds and tests zimcheck - among other configurations - on virtualized hardware too. The incurred slowdown results in the runtime exceeding the 0.5 second limit whereupon the performance report changes and the unit-tests fail (openzim/zim-tools#287).

The challenge was to come up with a solution meeting the following requirements:

a. no discrimination between fast and (moderately) slow build environments

b. the performance info is preserved in the output and is not excluded from comparison (so that, in the absence of dedicated
   performance testing, it keeps serving as a simple defence against unintended significant slowdown in zimcheck).

Since runtime numbers are mainly justified for large ZIM files, the solution is to report small runtimes as "<X seconds" for some value of X. The latter threshold was set to 3 with the only purpose of further increasing the amount of love in the world.

### DIFF
--- a/src/zimcheck/zimcheck.cpp
+++ b/src/zimcheck/zimcheck.cpp
@@ -349,7 +349,16 @@ int zimcheck (const std::vector<const char*>& args)
         const std::chrono::duration<double> runtime(endtime - starttime);
         const long seconds = lround(runtime.count());
         std::ostringstream ss;
-        ss << "[INFO] Total time taken by zimcheck: " << seconds << " seconds.";
+        ss << "[INFO] Total time taken by zimcheck: ";
+        if ( seconds < 3 )
+        {
+          ss << "<3";
+        }
+        else
+        {
+          ss << seconds;
+        }
+        ss << " seconds.";
         error.infoMsg(ss.str());
     }
     catch (const std::exception & e)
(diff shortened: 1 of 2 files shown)


legoktm:
Very cute :D I think this is reasonable, as a 3 second timeout on the test. I'll cherry-pick this into Debian and see how it goes.

legoktm (approved):
All the architectures built successfully in Debian with this patch, so good from my perspective. Thanks!

Issue #295, opened 2022-03-28
zimsplit leaves huge 16GB last file with warning on wikipedia_en_all_nopic_2022-01.zim


ballerburg9005:
When I use zimsplit on this 47GB Wikipedia file, it correctly creates X GB large chunks but then always leaves a 16GB large file at the end.

opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimaa
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimab
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimac
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimad
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimae
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimaf
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimag
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimah
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimai
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimaj
WARNING: Part /mnt/2/wikipedia_en_all_nopic_2022-01.zimaj is bigger that max part size. (16828436877>4000000000)
opening new file /mnt/2/wikipedia_en_all_nopic_2022-01.zimak
zimsplit --size=4000000000 /mnt/2/wikipedia_en_all_nopic_2022-01.zim  8.15s user 107.82s system 9% cpu 20:35.43 total

If I had to take a guess, it would seem like there is some kind of unknown reason for this last chunk being created at 16GB. But then again, since the point of zimsplit is primarily to make chunks in order to save on FAT32, it would totally defeat the purpose if just one file was left larger than 4GB, no matter what reasons existed.

It does not matter what options and sizes I run zimsplit with. It always creates this last file with 16GB.

Can someone explain this and how to work around it?

[...]

PR #296, opened 2022-03-29
disclaimer in --help that zimsplit leaves certain parts of file unsplit

ballerburg9005:
https://github.com/openzim/zim-tools/issues/295

### DIFF
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -172,6 +172,9 @@ class ZimSplitter
 
 static const char USAGE[] = R"(
     zimsplit splits smartly a ZIM file in smaller parts.
+    
+    zimsplit cannot split certain internal portions of ZIM files (search index) and will leave them with a 
+    warning in excess of the specified size. Manually splitting the search index renders search inoperable.
 
 Usage:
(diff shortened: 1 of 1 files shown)


mgautierfr:
In fact this is a bit more complex.
The zim contents are stored (and regrouped) in cluster (which may be compressed or not). At creation we try to keep cluster under a certain size but it is not always possible (if the content itself is bigger that the targeted cluster size)
zimsplit split the zim at cluster boundaries. This ensure us that a content is never cut in two different parts.
Such "big" contents can be:
- The fulltext index (most of the time)
- Any big content we add to the zim. Mostly videos but it can be anything else.

We (xapian) don't know how to read a xapian database split in two different files, so if we cut the database, we cannot use it.
But splitting other cluster (and content) may have another impact. For example, to read video on android, we pass the offset of the video in the zim file to the android video player. If the video is cut in two parts, the player may become buggy (I think that android application check that before. But this has to be confirmed).

But anyway, _if we have a politic to allow splitting such big cluster_ we should probably extend zimsplit to allow it to split in a middle of a too big cluster and avoid to the user to do another pass on the splitting.

PR #309, opened 2022-07-05
Finite task queue in zimcheck

### LINKED ISSUE #308: Limit task queue size in zimcheck
Before #194 zimchecks processed entries one after another. Support for multi-threading relies on a task queue where every entry from the ZIM file is treated as a separate task. The main thread populates the queue and worker threads process tasks from it (this scheme is used also when zimcheck is run in "single-threaded" mode which means that there is a single worker thread).

The current problem is that there is no limit on the task queue size, and the main thread keeps filling it until all entries have been pushed into the queue. For large ZIM files this results in a lot of memory being consumed by the task queue. For example, on a machine with 8GB of RAM `zimcheck --url_internal --url_external ` fails to handle [`stackoverflow.com_en_all_2022-05.zim`](http://download.kiwix.org/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim) (which has more than 64M entries).

veloman-yunkan:
Fixes #308

This PR was created on top of #307 in order to be able to test it on [stackoverflow.com_en_all_2022-05.zim](http://download.kiwix.org/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim). #307 must be merged first.


mgautierfr:
LGTM. Need #307 corrected and merged before this one.

mgautierfr (approved):

PR #316, opened 2022-08-14
Respecting the --ns option in zimdump show

### LINKED ISSUE #315: zimdump show weird behaviour
```
$ zimdump show --ns=M --url=Name gutenberg_fa_all_2022-01.zim 
Entry not found
```

is wrong, it should deliver the name like:
```
$ zimdump show --url=M/Name gutenberg_fa_all_2022-01.zim 
gutenberg_fa_all_2022-01
```

veloman-yunkan:
Fixes #315

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -103,6 +103,7 @@ class ZimDumper
     int listEntriesByNamespace(const std::string ns, bool details);
 
     zim::Entry getEntryByPath(const std::string &path);
+    zim::Entry getEntryByNsAndPath(char ns, const std::string &path);
     zim::Entry getEntry(zim::size_type idx);
 
     void dumpFiles(const std::string& directory, bool symlinkdump, std::function<bool (const char c)> nsfilter);
@@ -114,6 +115,11 @@ class ZimDumper
 zim::Entry ZimDumper::getEntryByPath(const std::string& path)
 {
     return m_archive.getEntryByPath(path);
+}
+
+zim::Entry ZimDumper::getEntryByNsAndPath(char ns, const std::string &path)
+{
+    return m_archive.getEntryByPathWithNamespace(ns, path);
 }
 
 zim::Entry ZimDumper::getEntry(zim::size_type idx)
@@ -380,7 +386,9 @@ Return value:
       See DIR/dump_errors.log for the listing of the errors.
 )";
 
-int subcmdInfo(ZimDumper &app, std::map<std::string, docopt::value> &args)
+typedef std::map<std::string, docopt::value> Options;
+
+int subcmdInfo(ZimDumper &app, Options &args)
 {
     app.printInfo();
     return 0;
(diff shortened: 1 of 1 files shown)


mgautierfr:
This PR add the `ns` to the path if `--ns` is provided.
It works and technically fix the behavior first explained by @kelson42 but it doesn't fix the global interface problem (and the solution proposed in https://github.com/openzim/zim-tools/issues/315#issuecomment-1235570245)

We can merge this PR anyway but not sure we should really close #315

kelson42:
@mgautierfr We need to know what is in the ZIM file. For the moment we can't properly because the "removal" of namespaces has broken `zimdump`. At this stage we are mostly interested to get the feature back before thinking of doing it better.

From a general POV, Libzim7 has introduced the idea that end-user visible content (so not the metadata, ...) would be "all" in the same namespace `C`. This mostly to simplify scrapers. That does not mean that we hide things so far that even an introspection tool like `zimdump` is not able *easily* to retrieve what is a ZIM file or rely on the concept of a namespace. [Namespace is a documented part of the ZIM specification](https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces).

If I agree to abstract/hide this concept for end-use I think I can pretty much say I disagree to hide it from the ZIM tools.

kelson42 (changes requested):
@veloman-yunkan Late, but this PR has not been abandoned!

I have tested:
```
$ zimdump show --ns=M --url=Name gutenberg_fa_all_2023-08.zim 
Entry not found
```

I don't know if I do something wrong but to me, it seems to fail (at least with latest libzim).

In addition, I have remarked that basically all the `--ns` options are broken. I would really like to have them working like announced and like they were. Today, I have basically lost more than one hour because - with `zimdump` - I was not even able to know if a ZIM file has a fulltext index or not! This is problematic.

veloman-yunkan:
> In addition, I have remarked that basically all the `--ns` options are broken. I would really like to have them working like announced and like they were. Today, I have basically lost more than one hour because - with `zimdump` - I was not even able to know if a ZIM file has a fulltext index or not! This is problematic.

It is possible to fix `zimdump` without/before addressing openzim/libzim#843, but should we do that?

mgautierfr:
> It is possible to fix zimdump without/before addressing https://github.com/openzim/libzim/issues/843, but should we do that?

Is the fix for https://github.com/openzim/libzim/issues/843 would invalidate the fix of zimdump ? If no, we should fix zimdump without waiting for libzim IMO.

veloman-yunkan:
> > It is possible to fix zimdump without/before addressing [openzim/libzim#843](https://github.com/openzim/libzim/issues/843), but should we do that?
> 
> Is the fix for [openzim/libzim#843](https://github.com/openzim/libzim/issues/843) would invalidate the fix of zimdump ? If no, we should fix zimdump without waiting for libzim IMO.

I was under the wrong impression (again mislead by the docstring) that `zim::Archive::findByPath()` respects the namespace and can be used instead of `zim::Archive::getEntryByPath()`. Or maybe I hoped to use `findx()`, having overlooked that it is a member function of `FileImpl` rather than `Archive`. In any case, that was a false call. :(

kelson42:
@veloman-yunkan I'm slightly lost about the status of this PR, but please let me know when this is ready to new functional review.

veloman-yunkan:
Now fixed taking advantage of `zim::Archive::getEntryByPathWithNamespace()` introduced in openzim/libzim#843

mgautierfr (approved):

kelson42:
I will check from a user perspective.

PR #321, opened 2022-09-07
New version 3.1.2

mgautierfr:


mgautierfr:
Packaging is broken. But we need to release anyway. We will fix the packaging and do another release later.

PR #348, opened 2023-05-16
feature: ZimMount

lyc8503:
Current progress:
- [x] Options to mount Zim files via command line interface
- [x] List files

Not working:
- [ ] Tests / CI are failing

### DIFF
--- a/src/zimmount/zimmount.h
+++ b/src/zimmount/zimmount.h
@@ -1,0 +1,38 @@
+#ifndef OPENZIM_ZIMMOUNT_H
+#define OPENZIM_ZIMMOUNT_H
+
+#include <string>
+#include <unordered_map>
+
+#define ZIM_PRIVATE
+#include <zim/archive.h>
+#include <zim/item.h>
+
+#define INVALID_INDEX 0xffffffff
+
+class ZimFSNode
+{
+  std::string name;
+
+ public:
+  zim::entry_index_type entry_index;
+  std::unordered_map<std::string, ZimFSNode*> children;
+
+  ZimFSNode();
+  ZimFSNode(const std::string& name, zim::entry_index_type index);
+
+  void addEntry(const zim::Entry& entry);
+  ZimFSNode* findPath(const std::string& path);
+  ~ZimFSNode();
+};
+
+class ZimFS
+{
+  zim::Archive archive;
+
+ public:
+  ZimFSNode root;
+  ZimFS(const std::string& fname, bool check_intergrity);
+};
+
+#endif  // OPENZIM_ZIMMOUNT_H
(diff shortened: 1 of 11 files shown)


src/zimmount/zimmount.h
```diff
@@ -0,0 +1,38 @@
+#ifndef OPENZIM_ZIMMOUNT_H
+#define OPENZIM_ZIMMOUNT_H
+
+#include <string>
+#include <unordered_map>
+
+#define ZIM_PRIVATE
```
mgautierfr, line 7: I don't think we need `ZIM_PRIVATE` here.
Using the public API of libzim should be enough.

src/zimmount/zimmount.h
```diff
@@ -0,0 +1,38 @@
+#ifndef OPENZIM_ZIMMOUNT_H
```
mgautierfr, line 1: Do we really need a `zimmount.h` ?
`zimmount` is the main binary, it cannot be "included" by other program.

But we can create a `zim_node.h`/`zim_node.cpp` as it will probably grow up with time.

mgautierfr (commented):
This first version seems good  (well, it is a first version :) )

While we are here, I want to mention a important thing:

A zim file can contains this four entries:
- `foo`
- `foo/`
- `foo/index.hmtl`
- `foo/bar.html`.

So we have two entries in the zim files that are also directory on the filesystem. We have to handle that correctly (ignore them ? Map them to something else ?) and not crash or hide children.

For now, we would create a `foo` node with a index.
Then we would create <code></code>, `index.html` and `bar.html`  nodes in `foo` node.
But as the `foo` node has a index, it will be shown as a file on filesystem and other entries will be hidden.

I think this issue is the real challenge here. All the others are pretty technical.

lyc8503:
> This first version seems good (well, it is a first version :) )
> 
> While we are here, I want to mention a important thing:
> 
> A zim file can contains this four entries:
> 
> * `foo`
> * `foo/`
> * `foo/index.hmtl`
> * `foo/bar.html`.
> 
> So we have two entries in the zim files that are also directory on the filesystem. We have to handle that correctly (ignore them ? Map them to something else ?) and not crash or hide children.
> 
> For now, we would create a `foo` node with a index. Then we would create , `index.html` and `bar.html` nodes in `foo` node. But as the `foo` node has a index, it will be shown as a file on filesystem and other entries will be hidden.
> 
> I think this issue is the real challenge here. All the others are pretty technical.

I took a look at the implementation of [zimdump](https://github.com/openzim/zim-tools/blob/main/src/zimdump.cpp#L258), and to my understanding, it would dump the file `foo` to `_exceptions/foo` and create `foo` folder (and its children) in its original place in such case.

I think I could try to implement the same behavior in zimmount.

PR #350, opened 2023-05-19
Meson using libmagic pkgconfig

### LINKED ISSUE #344: Use libmagic.pc for compilation
We use a somehow old way to detect magic library and it as been probably made this way as libmagic didn't provide a libmagic.pc at the time. Now it is the case so we would better fix the dependency to libmagic.

From #343

kelson42:
Related to #344

Unfortunately we can not fix it for the moment because Focal uses an old version (5.38) of libmagic and pkgconfig support has been added in 5.39.

But for all other system we use now the standard procedure and hopefuly we can remove the old custom code soon when Focal will move forward.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,10 +42,10 @@ jobs:
           - win32_dyn
         include:
           - target: native_static
-            image_variant: bionic
+            image_variant: focal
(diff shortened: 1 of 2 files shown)


kelson42:
Depends on https://github.com/kiwix/kiwix-build/pull/617

Issue #352, opened 2023-06-17
Incorrect `--illustration` image size lead to  SIGSEGV instead of a proper warning


shenlebantongying:
# Minimum demo
[zim_dict.zip](https://github.com/openzim/zim-tools/files/11780032/zim_dict.zip)
```
~/src/zim-tools/build/src/zimwriterfs/zimwriterfs  --welcome A/index.html --illustration=lenna_wrong.png --language=eng --title=zimdict --description=asd --longDescription=asdasd --creator=slbtty --publisher=slbtty --name=hwat ./zim_dict/ ./ok.zim
```

The crash exists in both the master as of today https://github.com/openzim/zim-tools/commit/6c1d2f5cf4e1a22ba9dab4f579da6d8c983f1adb and 3.2.0 installed from archlinux's repo.

Using `--illustration=lenna_wrong.png`, which has a size that is not 48x48 will lead to SIGEGV instead of printing a warning.

Using `--illustration=lenna_correct.png` will generate as usual.

# Expected

If the image size is wrong, it should print a warning instead of crashing.

# GDB

There is nothing related to image size in the backtrace. It seems to be stuck in an infinity loop.

```C++
Starting program: /home/slbtty/src/zim-tools/build/src/zimwriterfs/zimwriterfs --welcome=A/index.html --illustration=lenna_wrong.png --language=eng --title=zimdict --description=asd --longDescription=asdasd --creator=slbtty --publisher=slbtty --name=hwat ./zim_dict/ ./ok.zim
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x000055555557b47f in std::__detail::_Executor<__gnu_cxx::__normal_iterator<char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::__cxx11::sub_match<__gnu_cxx::__normal_iterator<char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >, std::__cxx11::regex_traits<char>, true>::_M_dfs (this=0x7fffffffcc80,
[...]

PR #353, opened 2023-06-18
Fix and enhance illustration file validation

shenlebantongying:
* Read the length & width directly from the png file to validate size.
* Replace the crazy regex with simpler magic number checking.

fix https://github.com/openzim/zim-tools/issues/352


mgautierfr:
We should probably use the "complex metadata check" (as it is done for long description at https://github.com/shenlebantongying/zim-tools/blob/fix_png/src/metadata_constraints.cpp#L28-L32) instead of moving the  check in `zimwriterfs`.

Else we will lost the check for other tools than zimwriterfs (zimcheck)

shenlebantongying:
I will redo this PR when I find time this month.

PR #356, opened 2023-06-21
[zimwriterfs] Use the correct method to add the illustration to zim file.

### LINKED ISSUE #355: Zimwriterfs create zim file with compressed illustration (wrong mimetype)
Zimwriterfs add all metadata using `addMetada` (https://github.com/openzim/zim-tools/blob/main/src/zimwriterfs/zimwriterfs.cpp#L126). But the default mimetype in this case is `text/plain;charset=utf-8` which is obviously wrong for illustration.

mgautierfr:
Fix #355

### DIFF
--- a/src/zimwriterfs/zimwriterfs.cpp
+++ b/src/zimwriterfs/zimwriterfs.cpp
@@ -123,7 +123,11 @@ bool checkMetadata(const zim::Metadata& metadata)
 void addMetadata(ZimCreatorFS& zimCreator, const zim::Metadata& metadata)
 {
   for ( const auto& kv : metadata ) {
-    zimCreator.addMetadata(kv.first, kv.second);
+    if (kv.first == "Illustration_48x48@1") {
+      zimCreator.addIllustration(48, kv.second);
+    } else {
+      zimCreator.addMetadata(kv.first, kv.second);
+    }
   }
 }
 


src/zimwriterfs/zimwriterfs.cpp
```diff
@@ -123,7 +123,11 @@ bool checkMetadata(const zim::Metadata& metadata)
 void addMetadata(ZimCreatorFS& zimCreator, const zim::Metadata& metadata)
 {
   for ( const auto& kv : metadata ) {
-    zimCreator.addMetadata(kv.first, kv.second);
+    if (kv.first == "Illustration_48x48@1") {
+      zimCreator.addIllustration(48, kv.second);
+    } else {
```
veloman-yunkan, line 128: It will take slightly more effort, but while we are at it, why not fix this issue for all illustrations of the NxN@1 kind?

mgautierfr: Well.. it is two different issues :)

Supporting NxN@1 is not just parse the "Illustration_NxN@1" string. This string is already fixed and is added by https://github.com/openzim/zim-tools/blob/main/src/zimwriterfs/zimwriterfs.cpp#L100-L103

We will have to extend the zimwriterfs options to allow user to pass other illustration other than 48x48 first (and if we do this, probably not pass through the `Metadata` class (or extend it to store the size as plain value)).

I not really sure we should do this in this PR which "simply" set the correct mimetype for illustration.

veloman-yunkan: This hack creates a situation where the same bug is likely to be reintroduced for illustrations with dimensions different from 48x48 when `zimwriterfs` is enhanced to support them. One way to track it is to add an action item (with a checkbox) to the corresponding ticket (or create such a ticket if one doesn't yet exist).

rgaudin: There's no such ticket ATM and given:
- we don't quite support other sizes (no reader makes use of them at least)
- we don't use zimwriterfs anymore ourselves for creating ZIM files
I'd say it's safe to merge as-is.

mgautierfr:
Packages CI is failing because of https://github.com/openzim/zim-tools/pull/362 merged a bit too soon.

I'm merging this PR as we seem to agree (at least, not counter argument since a month).
Issue https://github.com/openzim/zim-tools/issues/364 opened to follow the discussion.

PR #358, opened 2023-07-01
fix: rewrite illustration check to not use regex

shenlebantongying:
fix https://github.com/openzim/zim-tools/issues/352 by not using regex on binary data.

### DIFF
--- a/src/metadata_constraints.cpp
+++ b/src/metadata_constraints.cpp
@@ -21,7 +21,7 @@ const Metadata::ReservedMetadataTable reservedMetadataInfoTable = {
     MANDATORY,
     0, // There are no constraints on the illustration metadata size
     0, // in order to avoid decoding it as UTF-8 encoded text
-    PNG_REGEXP
+    ""
   },
 };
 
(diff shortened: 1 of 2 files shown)


src/metadata_constraints.cpp
```diff
@@ -30,3 +30,27 @@ METADATA_ASSERT("LongDescription shouldn't be shorter than Description")
...
+
+METADATA_ASSERT("Illustration must exist and be a 48x48 PNG file")
+{
+  if (!data.has("Illustration_48x48@1")) {
+    return false;
+  }
+
+  const std::string& illustration = data["Illustration_48x48@1"];
+
+  if (illustration.length() >= 24) {
```
veloman-yunkan, line 42: Why is this test needed? What's wrong with always (unconditionally) testing the first 8 bytes?

src/metadata_constraints.cpp
```diff
@@ -30,3 +30,27 @@ METADATA_ASSERT("LongDescription shouldn't be shorter than Description")
...
+    if (!(illustration.substr(0, 8) == "\x89\x50\x4e\x47\x0d\x0a\x1a\x0a")) {
+      return false;
+    }
+  }
+
+  // In PNG file, 16-20 is width and 20-24 is height in big endian order.
+  const std::string png48x48 = {0x00, 0x00, 0x00, 0x30, 0x00, 0x00, 0x00, 0x30};
+
+  if (illustration.substr(16, 8) != png48x48) {
```
veloman-yunkan, line 51: `illustration.substr(16, 8)` will throw if `illustration.length() < 16`

veloman-yunkan (changes requested):

kelson42:
@shenlebantongying Any news?

PR #359, opened 2023-07-04
--redirect: fix check for number of columns in redirect file

haoess:


kelson42:
Not sure exactly what are the rationals behind both bug and fix. But this probably sesserve a better automated testing.

rgaudin (approved):
Here's a valid redirect file

```
 \tAccueil\tBienvenue !!\twelcome\n
 \tAccueil2\t\tcommons48.png\n
 \timage\t\tcommons48.png\n
```

Notice that it starts with a space and a tab ` \t`. That space was used by the (now unused) namespace.

We have support for it in scraperlib [here](https://github.com/openzim/python-scraperlib/blob/main/src/zimscraperlib/zim/filesystem.py#L101) with [a test](https://github.com/openzim/python-scraperlib/blob/main/tests/zim/test_fs.py#L26)

As per [the doc](https://en.cppreference.com/w/cpp/regex/regex_search), `size()` returns _number of [marked subexpressions](https://en.cppreference.com/w/cpp/regex/ecmascript#Sub-expressions) plus 1_


[blame](https://github.com/openzim/zim-tools/blame/main/src/zimwriterfs/zimcreatorfs.cpp) clearly shows that @mgautierfr changed the regexp 3y ago to remove the namespace but did not adjust the match count. scraperlib's implementation precedes the change.

mgautierfr:
Thanks @rgaudin for the investigation.

> Here's a valid redirect file [...]

Is it really a valid file ? The regex search for 2 `\t` (3 fields) but you have 3 `\t`.

@haoess Please remove the merge commit of your own PR in your own `main`. 

@haoess It would be nice to have a testing of that code. Can you create it ? (It would probably need some refactoring of the code to separate the parsing of the file from adding the redirect)

haoess:
> @haoess Please remove the merge commit of your own PR in your own `main`.

(Hopefully) done.

> @haoess It would be nice to have a testing of that code. Can you create it ? 

Sorry, I’m not a C++ dev. Applying some knowledge about how other languages handle captures in RE, using a pattern with **three** pairs of parens `(…)` and testing the number of captures (+ 1 for the whole match) against **5** is not supposed to work (= what @rgaudin [says](https://github.com/openzim/zim-tools/pull/359#pullrequestreview-1528718606)).

PR #362, opened 2023-07-26
Add Debian to the CI

kelson42:
Add testing of Debian packaging in CI

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -15,6 +15,9 @@ jobs:
       fail-fast: false
       matrix:
         distro:
+          - debian-unstable
+          - debian-bullseye
+          - debian-buster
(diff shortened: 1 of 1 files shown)


mgautierfr:
Please don't merge PR when the CI don't pass !

mgautierfr:
Well, what ever is the cause, don't merge PR when the CI don't pass :)

Having a "accepted" CI not passing will make people not looking about the cause when they see a "red" check. This is the best way to merge new bug. The proof: You haven't check what was the cause and merge the PR with "another" bug.
None of the broken PRs (c++14 or other cause) should have been merged.

kelson42:
@mgautierfr You are right but now we have to fix the problems in the code in top prio on anything else, which is what I actually wanted as well.

PR #368, opened 2023-08-31
Move to c++17.

mgautierfr:
All our compilers should handle c++17. Let's move on.

Following of openzim/libzim#819
Fix openzim/libzim#757

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('zim-tools', ['c', 'cpp'],
   version : '3.2.0',
   license : 'GPLv3+',
-  default_options : ['c_std=c11', 'cpp_std=c++11', 'werror=true'])
+  default_options : ['c_std=c11', 'cpp_std=c++17', 'werror=true'])
 
 add_global_arguments(['-Werror', '-Wall'], language:'cpp')
 


veloman-yunkan (approved):
I did my best to review the changes carefully. To the best of my understanding the changes are limited to replacing a single character `1` with a single character `7`. Despite the seemingly minor modification its impact can be very consequential due to the changes in the language specification of C++17 relative to C++11. In particular:

- New rules concerning the [order of evaluation](https://en.cppreference.com/w/cpp/language/eval_order). Some of those enhancements of the specification shrink the realm of undefined behavior, possibly breaking the functionality of working (though buggy) code that accidentally took advantage of a particular implementation by a compiler of the old gap in the language standard.
- [Removal of trigraphs](https://en.cppreference.com/w/cpp/language/operator_alternative#Trigraphs_.28removed_in_C.2B.2B17.29).

- Removal of the increment operator for the `bool` type. This can result in changed semantics of template metaprogramming code snippets relying on SFINAE.

- etc, and so on, ...

Therefore these changes must be thoroughly tested before they can be allowed into the next release.

P.S. :smile:

Issue #371, opened 2023-10-03
zimwriterfs: created zim-file triggers download dialog when trying to open any extensionless html page in Kiwix Desktop


ilyaigpetrov:
I've created a small zim package that includes an extensionless welcome html page.
If I try to open it in `kiwix-desktop_x86_64_2.3.1-4.appimage`, it triggers a download dialog instead of rendering the page.

```sh-session
# ls example.com
hello.html  main_page  zim_favicon.png
# cat example.com/main_page 
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>TEST</title>
  </head>
  <body>
Hello from the main page.
  </body>
</html>
# zimwriterfs --welcome=main_page --illustration=zim_favicon.png --language=eng --title=example.com --description=nodescription --name=examplename -L longdescriptionhere --creator=https://github.com/ballerburg9005/wget-2-zim --publisher=wget-2-zim /root/wget-2-zim/example.com /root/wget-2-zim/example.com.zim
```
![zim-main-page-download-triggered](https://github.com/openzim/zim-tools/assets/1090811/91158d19-5097-4c83-8e77-d148dec78af1)

Expected: `main_page` must be rendered as html.
Actual: it is offered to be downloaded as file.

If I download https://www.mirrorservice.org/sites/download.kiwix.org/zim/wikipedia/wikipedia_ru_chemistry_mini_2023-08.zim and dump its content then I see it has extensionless html files such as `A/Яд` which are rendered correctly by Kiwix Desktop. I want to reproduce the same behavior for my zim-file.

My higher objective is to save a media-wiki site to a zim.

Thanks.

mgautierfr:
You are proposed to download the main_page because the mimetype is not supported.
The mimetype is set by zimwriterfs based on first the extension and then the content (if first method didn't help).

A `file -i main_page` gives a mimetype `text/html` so it should be good. I don't know what is the issue here.

> My higher objective is to save a media-wiki site to a zim.

You have [mwoffliner](https://github.com/openzim/mwoffliner) for that.

ilyaigpetrov:
> I don't know what is the issue here.

Could you, please, try reproducing the wrong behavior and decide if it's a bug or a mistake on my side?

ilyaigpetrov:
[example.com.zip](https://github.com/openzim/zim-tools/files/12794025/example.com.zip)

mgautierfr:
[...]

PR #372, opened 2023-10-06
Specifying MAGIC environment variable

rgaudin:
Added libmagic explicitly as a way to document that this is a required runtime dependency and because the `MAGIC` environ is meant for it but it was already included indirectly.

See #371


kelson42 (commented):
lgtm

PR #374, opened 2023-10-10
Fix zimwriterfs and missing magic file.

### LINKED ISSUE #371: zimwriterfs: created zim-file triggers download dialog when trying to open any extensionless html page in Kiwix Desktop

mgautierfr:
Fix #371

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,17 @@ distribution. Try then with a source tarball distributed by the
 problematic upstream project or even directly from the source code
 repository.
 
+### Magic File
+
+`zimwriterfs` use libmagic to detect the mimetype of files to add.
+It needs to load the magic database. Most of the time, this file is already present
+in your system and everything works.
+But sometime, this file may be not present or cannot be found.
+You can set the `MAGIC` environment variable to point to this file.
+
+`zimwriterfs` will refuse to run if this file is not found. You can force
+it to run passing the option `--skip-libmagic-check`.
+
 License
 -------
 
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
LGTM but I notice that `parse_args` was inside a try block and is not anymore.

mgautierfr:
Yes. But for me it should not be in. I suspect a too "large" try block.
The `try/catch` is to catch error from the zim creation. `parse_args` has no reason to throw a exception.

PR #376, opened 2023-10-20
Remove usage of deprecated `std::codecvt_utf8_utf16`.

mgautierfr:
`std::codecvt_utf8_utf16` is deprecated in c++17.
We replace it with the usage of windows' `MultiByteToWideChar` function.

The new `converter` method is taken from libzim's `FS::toWideChar` in `src/fs_windows.cpp` with the only difference to return `std::wstring` instead of `std::unique_ptr<wchar_t[]>`. (https://github.com/openzim/libzim/blob/main/src/fs_windows.cpp#L131-L144)

(Triggered by https://github.com/kiwix/kiwix-build/actions/runs/6575114946/job/17861540481?pr=638)

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -31,7 +31,6 @@
 #include <sys/stat.h>
 #include <iomanip>
 #include <vector>
-#include <codecvt>
 #include <unordered_map>
 
 #include "version.h"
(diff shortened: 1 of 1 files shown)


veloman-yunkan (changes requested):

src/zimdump.cpp
```diff
@@ -49,7 +48,22 @@
...
 
-std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+
+#ifdef _WIN32
+std::wstring utf8ToUtf16(const std::string& string) {
+  auto size = MultiByteToWideChar(CP_UTF8, 0,
+                path.c_str(), -1, nullptr, 0);
+  auto wdata = std::wstring(size, 0);
+  auto ret = MultiByteToWideChar(CP_UTF8, 0,
+                path.c_str(), -1, wdata.get(), size);
```
veloman-yunkan, line 58: In [my previous review comment](https://github.com/openzim/zim-tools/pull/376#discussion_r1367015751), the main concern was that - as a result of copy-paste-modify - the variable name  `path` became out of sync with the function parameter name (which was changed to `string`). After you fix it and the CI win32 build passes there is no need to request another review.

veloman-yunkan (changes requested):

Issue #377, opened 2023-10-24
Correctly handle alias/hardlink in content duplication check.


mgautierfr:
Following https://github.com/openzim/libzim/issues/824 we will need to extend check tools to correctly handle duplication.

- Duplication at entry level (two entries with same content) is valid
- Duplication at blob level (two blobs with the same content) is not valid.

It implies we loop on the blobs instead of the entries to check for duplicated level.
And if we do so, we may want to do other checks at this moment instead of looping on articles after.

PR #379, opened 2023-10-30
Do not detect clone entry as duplicated content.

mgautierfr:
If two entries point to the same tuple (cluter_id, blob_id), we must not report them as duplicated content.

Depends of https://github.com/openzim/libzim/pull/833

Missing: Test of the check with a zim file actually containing clone entries.

### DIFF
--- a/src/zimcheck/checks.cpp
+++ b/src/zimcheck/checks.cpp
@@ -18,27 +18,6 @@
 #include <zim/archive.h>
 #include <zim/item.h>
 
-// Specialization of std::hash needed for our unordered_map. Can be removed in c++14
-namespace std {
-  template <> struct hash<LogTag> {
-    size_t operator() (const LogTag &t) const { return size_t(t); }
-  };
-}
-
-// Specialization of std::hash needed for our unordered_map. Can be removed in c++14
-namespace std {
-  template <> struct hash<TestType> {
-    size_t operator() (const TestType &t) const { return size_t(t); }
-  };
-}
-
-// Specialization of std::hash needed for our unordered_map. Can be removed in c++14
-namespace std {
-  template <> struct hash<MsgId> {
-    size_t operator() (const MsgId &msgid) const { return size_t(msgid); }
-  };
-}
-
 namespace
 {
 
--- a/src/zimcheck/checks.cpp
+++ b/src/zimcheck/checks.cpp
@@ -14,6 +14,7 @@
 #include <mutex>
 #include <thread>
 #include <queue>
+#include <optional>
 #include <zim/archive.h>
 #include <zim/item.h>
 
@@ -111,6 +112,11 @@ typedef std::map<MsgParams::key_type, MsgParams::mapped_type> SortedMsgParams;
 SortedMsgParams sortedMsgParams(const MsgParams& msgParams)
 {
   return SortedMsgParams(msgParams.begin(), msgParams.end());
+}
+
+bool areAliases(const zim::Item& i1, const zim::Item& i2)
+{
+    return i1.getClusterIndex() == i2.getClusterIndex() && i1.getBlobIndex() == i2.getBlobIndex();
 }
 
 } // unnamed namespace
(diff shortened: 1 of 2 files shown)


src/zimcheck/checks.cpp
```diff
@@ -387,7 +387,7 @@ void ArticleChecker::check_item(const zim::Item& item)
         data = item.getData();
 
     if(checks.isEnabled(TestType::REDUNDANT))
-        hash_main[adler32(data)].push_back( item.getIndex() );
+        hash_main[adler32(data)].push_back( {item.getIndex(), item.getClusterIndex(), item.getBlobIndex()} );
```
veloman-yunkan, line 391: Why storing the cluster and blob indices with the entry index (and increasing the memory usage by zimcheck) if that information can be obtain in `ArticleChecker::detect_redundant_articles()`?

src/zimcheck/checks.cpp
```diff
@@ -487,14 +487,19 @@ void ArticleChecker::detect_redundant_articles()
         progress.report();
         auto l = it.second;
         while ( !l.empty() ) {
-            const auto e1 = archive.getEntryByPath(l.front());
+            const auto [e1_idx, e1_cluster_idx, e1_blob_idx] = l.front();
             l.pop_front();
+            const auto e1 = archive.getEntryByPath(e1_idx);
             if ( !l.empty() ) {
                 // The way we have constructed `l`, e1 MUST BE an item
                 const std::string s1 = e1.getItem().getData();
```
veloman-yunkan, line 495: Before aliases were introduced, this extra access to the item data happened only for the hopefully rare false positives due to hash conflicts. Now the item data will be loaded during the redundant item check for each entry having an alias. If aliases are going to be used mainly for large binary content (e.g. youtube videos) then we better have this access optimized.

veloman-yunkan (commented):
From the functionality point of view it should work, but I have a couple of comments regarding the performance impact of this fix.

kelson42:
@mgautierfr Now that libzim `9.1.0` has been released, we should finish/merge this PR.

src/zimcheck/checks.cpp
mgautierfr: It was to avoid getting the dirent again in `detect_redundant_articles()`.
But, indeed, it may not worth it. I have changed that.

src/zimcheck/checks.cpp
mgautierfr: Indeed. Done.

src/zimcheck/checks.cpp
```diff
@@ -487,15 +488,26 @@ void ArticleChecker::detect_redundant_articles()
...
+                std::optional<std::string> s1;
                 decltype(l) articlesDifferentFromE1;
                 for(auto other : l) {
-                    auto e2 = archive.getEntryByPath(other);
-                    std::string s2 = e2.getItem().getData();
+                    // The way we have constructed `l`, e2 MUST BE an item
+                    const auto e2 = archive.getEntryByPath(other).getItem();
+                    const auto e2_cluster_idx = e2.getClusterIndex();
+                    const auto e2_blob_idx = e2.getBlobIndex();
+                    if (e1_cluster_idx == e2_cluster_idx && e1_blob_idx == e2_blob_idx) {
```
veloman-yunkan, line 504: I'd like to have here a self-explaining `if ( areAliases(e1, e2) )`  and less clutter in this function due to the extra variables `e?_cluster_idx` and `e?_blob_idx`.

veloman-yunkan (changes requested):
Please rebase-fixup while addressing the last comment and the PR can be merged.

mgautierfr:
Last comment fixed.
Commits are already fixed up.
New (last) commit remove `std::hash` specialization as we are now > c++14.

veloman-yunkan (approved):

mgautierfr:
CI for packages on debian unstable is failing because of a wrong version of libzim available.
I'm merging this issue as it is not related (or it is related, but it's not here that we will fix this)

mgautierfr:
Fix #377

PR #382, opened 2023-11-01
Multiple CI workflow improvements

kelson42:
* Use latest version of kiwix-build CI images
* Add macos12 and macos13 to CI
* Use latest version of `actions/setup-python` and Python 3.9 (in place of 3.7) for macOS


kelson42:
@mgautierfr I don't get why it does not compile on macos12 and macos13. Do you? Why?

mgautierfr (approved):

PR #400, opened 2024-04-13
ZIM Fuse Filesystem

### LINKED ISSUE #79: ZIM Filesystem Fuse Module
This ticket tracks the [Kiwix GSoC 2023 Project](https://www.kiwix.org/en/support-us/code/google-summer-of-code/) **ZIM Filesystem Fuse Module** until code contributions and/or specific tickets requires creating its own dedicated repository (or falls into [zim-tools](https://github.com/openzim/zim-tools)).

**Candidates, contributors**, this ticket is the preferred location to discuss this project. Ask your questions here and mentors (@mgautierfr, @kelson42) shall respond.

**Mandatory reads**:
- https://github.com/kiwix/overview/blob/main/CONTRIBUTING.md
- https://github.com/kiwix/overview/blob/main/CODE_OF_CONDUCT.md

---
## ZIM Filesystem Fuse Module

**Objective**: we need to create a filesystem fuse module that enables access to the content of a ZIM file, allowing users to view entries as files without using zimdump.

**Technologies**: C++, Linux internals, FUSE

**Description**:

Kiwix provides offline access to Wikipedia and other educational content through its ZIM file format. Inspecting ZIM files is very useful for developers and ZIM creators. While the zimdump tool exists, it is not as convenient and easy to use as a filesystem. Therefore, we want to make it easier for users to access the contents of a ZIM file by creating a (read-only) filesystem fuse module.

The ZIM filesystem fuse module will be written in C++ and will use the libzim and FUSE library to enable access to the contents of a ZIM file as if it were a regular (yet read-only)  filesystem. The module will allow users to view the ZIM entries as a  tree or folders and files, the latter being readable as regular ones. This will make it easier for users to access the content of a ZIM file and will provide a more user-friendly interface for exploring its contents.

**Key Deliverables**:

- A C++ implementation of the ZIM filesystem fuse module.
- Documentation for how to install and use the module.
- Test cases to verify the correct functioning of the module.

Skills required:

- Strong knowledge of C++
- Familiarity with Linux internals, specifically the FUSE library
- Knowledge of filesystems and file access methods.

**Difficulty**: Hard. Expect 350 hours of work.

juuz0:
Fixes https://github.com/kiwix/overview/issues/79

**Mount any ZIM file to your filesystem**
Usage:
`zimfuse zimFile.zim mountDir`

mountDir should exist before using zimfuse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ Most famous tools are:
   opened with a ZIM reader; [Kiwix](https://kiwix.org) is one example,
   but there are [others](https://openzim.org/wiki/ZIM_Readers).
 
+* `zimfuse` creates a filesystem FUSE module that enables access to 
+  the content of a [ZIM](https://openzim.org) file, allowing users to 
+  view entries as files.
+
 A few other tools are provided as well but are of less value for most
 of the usages.
 
@@ -60,6 +64,8 @@ need to be available:
 * [Z](https://zlib.net/) (package `zlib1g-dev` on Debian/Ubuntu)
 * [Gumbo](https://github.com/google/gumbo-parser) (package `libgumbo-dev` on Debian/Ubuntu)
 * [ICU](http://site.icu-project.org/) (package `libicu-dev` on Debian/Ubuntu)
+
+`zimfuse` relies on [FUSE](https://github.com/libfuse/libfuse) (package `fuse3` on Debian/Ubuntu)
 
 These dependencies may or may not be packaged by your operating
 system. They may also be packaged but only in an older version. The
--- a/meson.build
+++ b/meson.build
@@ -29,7 +29,7 @@ icu_dep = dependency('icu-i18n', static:static_linkage)
 
 with_writer_and_mount = host_machine.system() != 'windows'
 
-if with_writer
+if with_writer_and_mount
   thread_dep = dependency('threads')
   zlib_dep = dependency('zlib', static:static_linkage)
   gumbo_dep = dependency('gumbo', static:static_linkage)
--- a/meson.build
+++ b/meson.build
@@ -27,13 +27,14 @@ rt_dep = dependency('rt', required:false)
 docopt_dep = dependency('docopt', static:static_linkage)
 icu_dep = dependency('icu-i18n', static:static_linkage)
 
-with_writer = host_machine.system() != 'windows'
+with_writer_and_mount = host_machine.system() != 'windows'
 
 if with_writer
   thread_dep = dependency('threads')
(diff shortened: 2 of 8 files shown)


kelson42:
I guess this would be an implementation of https://github.com/kiwix/overview/issues/79 ?

juuz0:
> I guess this would be an implementation of https://github.com/kiwix/overview/issues/79 ?

Yes.

kelson42:
@juuz0 Thanks but most of the CI is broken? Can you fix it?

juuz0:
CI fails because 'fuse3' is not available on the system...maybe something to add in kiwix-build later?

mgautierfr (changes requested):
Thanks for the PR @juuz0.
But I have few general comments better addressed on the full PR than each commit:

- This kind of tree is a perfect use case for OOP and inheritance.
  A basic `Node` with `name`, `parent`. Then a `Dir` node (inheriting `Node`) with `children` and a `Leaf` node.
- You define a `Node::Ptr` as `unique_ptr`, but the `Node` tree is composed of `Node*`. `Node::Ptr` is only used in `Tree::mappedNodes`. This make `Tree` the direct owner of all the nodes and `Nodes` only making "reference" to other nodes. It would make more sense to have `Dir` nodes the owners of their children.
- The `Tree::mappedNodes` is kind of global cache. But I'm not sure we need it. Either "directories" contains few entries and loop in the `vector` should be enough or not, and in this case, it would be better to transform `Node::children` into a `map`. We would have the O(1) access at node level and top of ensuring us the unity of children names.
- The `Tree::statCache` could be removed by simply move the cached `struct stat` in the `Node` itself.
- Then the `Tree` could probably be removed as it is simply a `Dir` node without parent.
- The `collisionCount` is local to each node but "global" to all children. This means that four twins `foo` (x2) and `bar` (x2), you will find the sanitized names `foo`, `foo(1)`, `bar` and `bar(2)`. It should be `bar(1)`.
- Redirects are not properly handle. You resolve the redirection and return the content of the target but it may break relative links in html. You should treat redirect as symlink (and so implement `readlink`)

On top of that, I wonder about the memory usage of the tree.
wikipedia_fr_all_maxi contains 7029908 entries (leafs). And `size_of(Node)` is 144 (without counting the actual data store in it (names/fullPath/originalPath bytes, children ptr). So it means that we need at least 965MiB for the nodes only. Probably more that 2GiB if we add the path and children ptr.
We can reduce that by carefully define `Node` structure and what we store but at the end, we will always use a lot of data.
(We may simply don't care and tell user that mounting zim files need a lot of memory, at least for now)

PR #411, opened 2024-05-27
--longDescription is not mandatory

### LINKED ISSUE #408: LongDescription incorrectly marked as mandatory
`zimwriterfs` usage:

```
Usage: zimwriterfs [mandatory arguments] [optional arguments] HTML_DIRECTORY ZIM_FILE

Purpose:
	Packing all files (HTML/JS/CSS/JPEG/WEBM/...) belonging to a directory in a ZIM file.

Mandatory arguments:
	-w, --welcome		path of default/main HTML page. The path must be relative to HTML_DIRECTORY.
	-I, --illustration	path of ZIM file illustration. The path must be relative to HTML_DIRECTORY and the image a 48x48 PNG.
	-l, --language		language code of the content in ISO639-3
	-t, --title		title of the ZIM file
	-d, --description	short description of the content
	-L, --longDescription	long description of the content
	-c, --creator		creator(s) of the content
	-p, --publisher		creator of the ZIM file itself

	HTML_DIRECTORY		path of the directory containing the HTML pages you want to put in the ZIM file.
	ZIM_FILE		path of the ZIM file you want to obtain.

Optional arguments:
	-v, --verbose		print processing details on STDOUT
	-h, --help		print this help
	-V, --version		print the version number
	-m, --clusterSize	number of bytes per ZIM cluster (default: 2048Kb)
	-J, --threads		count of threads to utilize (default: 4)
	-x, --inflateHtml	try to inflate HTML files before packing (*.html, *.htm, ...)
	-r, --redirects		path to a TSV file containing a list of redirects (url title target_url).
	-j, --withoutFTIndex	don't create and add a fulltext index of the content to the ZIM.
	-a, --tags		tags - semicolon separated
	-e, --source		content source URL
	-n, --name		custom (version independent) identifier for the content
	-o, --flavour		custom (version independent) content flavour
	-s, --scraper		name & version of tool used to produce HTML content
	--skip-libmagic-check	Accept to run even if magic file cannot be loaded (mimetypes in the zim file may be wrong).

Example:
	zimwriterfs --welcome=index.html --illustration=illustration.png --language=fra --title=foobar --description=mydescription \
		--creator=Wikipedia --publisher=Kiwix ./my_project_html_directory my_project.zim

Documentation:
	zimwriterfs source code: https://github.com/openzim/zim-tools
	ZIM format: https://openzim.org
```

`LongDescription` metadata is set via `--longDescription` which is marked as mandatory while it probably should not…
Actually, the flag is not checked for being used but if not set or set empty:

```
Metadata doesn't meet the following requirements:
    LongDescription shouldn't be shorter than Description
```

To me this is an incorrect interpretation of the spec which says _“extended description of content”_ not _alternative longer description_. The fact that it should be longer than description is preposterous.

kelson42:
Fixes #408

### DIFF
--- a/src/zimwriterfs/zimwriterfs.cpp
+++ b/src/zimwriterfs/zimwriterfs.cpp
@@ -76,6 +76,7 @@ bool thereAreMissingArguments()
     return false;
 
   return creator.empty()
+      || name.empty()
       || publisher.empty()
       || description.empty()
       || language.empty()
--- a/src/zimwriterfs/zimwriterfs.cpp
+++ b/src/zimwriterfs/zimwriterfs.cpp
@@ -285,7 +285,7 @@ void parse_args(int argc, char** argv)
 
   do {
     c = getopt_long(
-        argc, argv, "ahVvijxuw:I:t:d:c:l:p:r:e:n:m:J:UBL:", long_options, &option_index);
+        argc, argv, "a:hVvijxuw:I:t:d:c:l:p:r:e:n:m:J:UBL:", long_options, &option_index);
 
     if (c != -1) {
       switch (c) {
--- a/src/zimwriterfs/zimwriterfs.cpp
+++ b/src/zimwriterfs/zimwriterfs.cpp
@@ -186,8 +186,6 @@ void usage()
   std::cout << "\t-t, --title\t\ttitle of the ZIM file" << std::endl;
   std::cout << "\t-d, --description\tshort description of the content"
             << std::endl;
-  std::cout << "\t-L, --longDescription\tlong description of the content"
-            << std::endl;
   std::cout << "\t-c, --creator\t\tcreator(s) of the content" << std::endl;
   std::cout << "\t-p, --publisher\t\tcreator of the ZIM file itself"
             << std::endl;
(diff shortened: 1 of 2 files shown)


rgaudin:
I doubt this will fix #408 ; it's not just a usage issue

kelson42:
@rgaudin It is indeed not just a usage() change

src/zimwriterfs/zimwriterfs.cpp
```diff
@@ -283,7 +283,7 @@ void parse_args(int argc, char** argv)
 
   do {
     c = getopt_long(
-        argc, argv, "hVvijxuw:I:t:d:c:l:p:r:e:n:m:J:UBL:", long_options, &option_index);
+        argc, argv, "hVvijxuUBLw:I:t:d:c:l:p:r:e:n:m:J:", long_options, &option_index);
```
mgautierfr, line 286: Why changing this line ?
The `:` after the `L` was about `-L` (`--longDescription`) expecting a value.
If you remove it, getopt will probably not parse the option correctly.

mgautierfr (changes requested):

src/zimwriterfs/zimwriterfs.cpp
kelson42: Yes, will fix

PR #413, opened 2024-06-10
[CHECK] Do not continue with other checks if integrity check fails.

mgautierfr:
It make no sense to check content in a zim archive if the zim archive itself is broken and we can't read it.

Fix openzim/libzim#893

### DIFF
--- a/src/zimcheck/zimcheck.cpp
+++ b/src/zimcheck/zimcheck.cpp
@@ -266,70 +266,75 @@ int zimcheck (const std::vector<const char*>& args)
         error.infoMsg("[INFO] Zimcheck version is " + std::string(VERSION));
 
         //Test 0: Low-level ZIM-file structure integrity checks
-        if(enabled_tests.isEnabled(TestType::INTEGRITY))
-            test_integrity(filename, error);
-
-        // Does it make sense to do the other checks if the integrity
-        // check fails?
-        zim::Archive archive( filename );
-        error.addInfo("file_uuid",  stringify(archive.getUuid()));
-
-        //Test 1: Internal Checksum
-        if(enabled_tests.isEnabled(TestType::CHECKSUM)) {
-            if ( enabled_tests.isEnabled(TestType::INTEGRITY) ) {
-                error.infoMsg(
-                    "[INFO] Avoiding redundant checksum test"
-                    " (already performed by the integrity check)."
-                );
-            } else {
-                test_checksum(archive, error);
-            }
+        bool should_run_full_test = true;
+        if(enabled_tests.isEnabled(TestType::INTEGRITY)) {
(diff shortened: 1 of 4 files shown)


src/zimcheck/zimcheck.cpp
```diff
@@ -266,70 +266,73 @@ int zimcheck (const std::vector<const char*>& args)
...
+        if(enabled_tests.isEnabled(TestType::INTEGRITY)) {
+            valid = test_integrity(filename, error);
+        }
+
 
-        // Does it make sense to do the other checks if the integrity
-        // check fails?
-        zim::Archive archive( filename );
-        error.addInfo("file_uuid",  stringify(archive.getUuid()));
+        if (valid) {
```
veloman-yunkan, line 275: You may get here with a garbage value of `valid` when the integrity check is excluded (e.g. when running a single check). After you fix it by initializing the variable to `true` it may make sense to rename it.

Also it may make sense to output an explicit message when the integrity check is skipped. Something like:

> WARNING: Integrity check is skipped. Any detected errors (in case zimcheck doesn't just crash) may in fact be due to corrupted/invalid data.

veloman-yunkan (changes requested):

src/zimcheck/zimcheck.cpp
mgautierfr: Fixed (in the same commit).

I haven't keep the "(in case zimcheck doesn't just crash)" as it would also be a bug in libzim/zimcheck itself as we should not crash in case of invalid zim file.
Better have a issue opened in this case and not user thinking it is its zim file fault and go on.

src/zimcheck/zimcheck.cpp
```diff
@@ -266,70 +266,75 @@ int zimcheck (const std::vector<const char*>& args)
         error.infoMsg("[INFO] Zimcheck version is " + std::string(VERSION));
 
         //Test 0: Low-level ZIM-file structure integrity checks
-        if(enabled_tests.isEnabled(TestType::INTEGRITY))
-            test_integrity(filename, error);
+        bool should_run_full_test = true;
+        if(enabled_tests.isEnabled(TestType::INTEGRITY)) {
+            should_run_full_test = test_integrity(filename, error);
+        } else {
+            error.infoMsg("[WARNING] Integrity check is skipped. Any detected errors may in fact due to corrupted/invalid data.");
```
veloman-yunkan, line 273: ... in fact **be** due to ...

veloman-yunkan (commented):
This is technically an approval. Please fix the message before merging.

mgautierfr:
Fixed. Directly merge if you agree.

veloman-yunkan (approved):

Issue #414, opened 2024-06-14
zimdump dump misses html files when file name conflict with directory name


nickhuang99:
Let's use "real" wiki page of "C++" as an example:
"https://en.wikipedia.org/wiki/C++" is a html page and it has some sub pages under the directory "C++":
https://en.wikipedia.org/wiki/C++/CLI
This situation cannot be represented in dump static html files because "C++" cannot be a html file and directory at same time.
How should zimdump generate redirect is not so easy, especially when "C++" is url-escaped as "C%2B%2B". Then redirect URL has to include a safe-URL encoded to check if actual filesystem directory "C++" actually exists. 
To give a real test case, please download "kiwix" computer zim from:
https://download.kiwix.org/zim/wikipedia/wikipedia_en_computer_maxi_2024-05.zim

And when you zimdump it, you will see "C++" html page is missing because "C++" is a directory to hold "CLI" page in filesystem.

nickhuang99:
After submitted, I realized this is a duplicate of old issue https://github.com/openzim/zim-tools/issues/190
Can someone dup it or should I close it? Maybe just another test cases in future?

nickhuang99:
[escapeSlash.diff.zip](https://github.com/user-attachments/files/15878309/escapeSlash.diff.zip)
I have a simple solution to solve this issue by escaping all '/' in path of filename to allow all article/picture residing in same level of directory to avoiding this directory-filename-conflicts. Even though this may create potential filename longer than 255 issue when directory is too deep. However, my tests show at least for "https://download.kiwix.org/zim/wikipedia/wikipedia_en_computer_maxi_2024-05.zim" we no longer have any single exception.

Attached please find patch. Can developer  to take a look to see if it can be patched if ok.
Thank you.

PR #415, opened 2024-06-18
Escape all slashes in filename so all files reside in same level

nickhuang99:
There is some builtin conflict of directory and filename which is showed in [this issue](https://github.com/openzim/zim-tools/issues/414#issue-2354229961). So, the simple solution is just avoiding create directory which conflicts with filename, instead just "flatten" all filename by escape all slashes('/') to form all unique filenames.

For example, there is conflict of "C++" as both directory and filename for two pages:
C++
C++/CLI
Now two files: 
C++   
C++%2FCLI

My test shows at least for kiwix computer zim files, there is no single exceptions.

The only potential issue is when filename is more than 255 limit and that is why we need sub directories. However, this chances maybe not so much?

### DIFF
--- a/src/zimdump.cpp
+++ b/src/zimdump.cpp
@@ -291,6 +291,12 @@ void ZimDumper::writeHttpRedirect(const std::string& directory, const std::strin
     write_to_file(directory + SEPARATOR, outputPath, content.c_str(), content.size());
 }
 
+static void replaceSlash(std::string& filename) {
+  std::string::size_type p;
+  while ((p = filename.find('/')) != std::string::npos)
+    filename.replace(p, 1, "%2F");
+}
+
 void ZimDumper::dumpFiles(const std::string& directory, bool symlinkdump, std::function<bool (const char c)> nsfilter)
 {
   unsigned int truncatedFiles = 0;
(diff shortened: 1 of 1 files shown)


veloman-yunkan:
I don't agree with such an approach. The main problem is that it breaks the links (in the example above, a link to "C++/CLI" will stop working).

mgautierfr:
Agree with @veloman-yunkan.
Especially has we have made the move in the opposite direction few years ago : https://github.com/openzim/zim-tools/pull/85

We should better fix the writing of file/directory than going back to a flat dump. (At least we must discuss this on the functional level before doing it to fix an issue)

nickhuang99:
> I don't agree with such an approach. The main problem is that it breaks the links (in the example above, a link to "C++/CLI" will stop working).

You are absolutely correct. The internal URL won't work for this change. 


> Agree with @veloman-yunkan. Especially has we have made the move in the opposite direction few years ago : #85
> 
> We should better fix the writing of file/directory than going back to a flat dump. (At least we must discuss this on the functional level before doing it to fix an issue)

Thank you so much for pointing out this old history and I can see now it is just repeating and theoretically there is no solution for this kind of exception for dumping as static html files.

I decide this is not a solution at all. Please close this PR. Many thanks.

PR #417, opened 2024-06-21
using filesystem read_link to test if sym link already created.

nickhuang99:
1. using std::filesystem's read_link.
2. This is an improvement so only test when failure to avoid expensive system call.
3. Hopefully this avoid race condition.


src/zimdump.cpp
```diff
@@ -291,6 +292,16 @@ void ZimDumper::writeHttpRedirect(const std::string& directory, const std::strin
     write_to_file(directory + SEPARATOR, outputPath, content.c_str(), content.size());
 }
 
+inline static bool testSymLink(const std::string& sym, const std::string& target) {
```
mgautierfr, line 295: `testSymlink` convey very few information. (What is tested ?)

It would be better to rename it to `checkSymlinkTarget` or `symlinkPointsTo`.

mgautierfr (changes requested):
On code side, only a small change needed.

On the functionality itself, I'm a bit reluctant as we should not have this kind of situation. If symlinks already exist, there is a lot of probability that contents already exist too. Only checking for symlink is useless here. And if we overwrite content/symlink we should probably have a option at command line level to allow it.

But anyway, even with my concerns, I don't see how this can harm so I will not block this PR.

kelson42:
@nickhuang99 Have you abandoned this PR? Can you please answer feedback from @mgautierfr ?

PR #425, opened 2024-08-26
Introduce MS Windows CI

mgautierfr:


kelson42:
Considering that writer is available everywhere in libzim, why do we still have anything "without writer"?

mgautierfr:
This is zimwriterfs. We don't have libmagic on Windows.
The first commit of this PR (so the name) was about deactivating tests associated to zimwriterfs (and needed dependencies) when trying to compile without it (on Windows).

Since then, compilation of zim-tools on Windows was a bit more complex than expected so the PR is also more complex.

mgautierfr:
This Pr need port of zimcheck to docoptcpp to be "finished". (Will do)

@veloman-yunkan Please can you have a look to [`5186d50` (#425)](https://github.com/openzim/zim-tools/pull/425/commits/5186d505aaebe58453cb46283c5c0062ec0c7656)
I don't have a damn idea why we have this behavior.

src/metadata.cpp
```diff
@@ -48,7 +48,21 @@ bool searchRegex(const std::string& regexStr, const std::string& text)
...
+  //  : ptr_(str.data()), length_(static_cast<int32_t>(str.size())) { }
+  // StringPiece(const char* offset, int32_t len) : ptr_(offset), length_(len) { }
+  // ```
+  // However using the first constructor ends with a corrupted StringPiece (wrong ptr)
+  // and using second one works. Don't ask me why
+  // This is broken
+  // icu::StringPiece stringPiece(utf8EncodedString);
+  // This is not
+  icu::StringPiece stringPiece(utf8EncodedString.data(), static_cast<int32_t>(utf8EncodedString.size()));
+  return icu::UnicodeString::fromUTF8(stringPiece).length();
```
veloman-yunkan, line 65: The only explanation that I could find is because of the following template constructor of `StringPiece`:

```c++
  /**
   * Constructs from some other implementation of a string piece class, from any
   * C++ record type that has these two methods:
   *
   * \code{.cpp}
   *
   *   struct OtherStringPieceClass {
   *     const char* data();  // or const char8_t*
   *     size_t size();
   *   };
   *
   * \endcode
   *
   * The other string piece class will typically be std::string_view from C++17
   * or absl::string_view from Abseil.
   *
   * Starting with C++20, data() may also return a const char8_t* pointer,
   * as from std::u8string_view.
   *
   * @param str the other string piece
   * @stable ICU 65
   */
  template <typename T,
            typename = typename std::enable_if<
                (std::is_same<decltype(T().data()), const char*>::value
#if defined(__cpp_char8_t)
                    || std::is_same<decltype(T().data()), const char8_t*>::value
#endif
                ) &&
                std::is_same<decltype(T().size()), size_t>::value>::type>
  StringPiece(T str)
      : ptr_(reinterpret_cast<const char*>(str.data())),
        length_(static_cast<int32_t>(str.size())) {}
```

If overload resolution for some reason chooses that constructor over `StringPiece(const std::string& str)` then the string is passed into it by value, and the pointer is bound to the data of a temporary object that is destroyed after the completion of the `StringPiece` constructor. That explanation is valid for an explicit `stringPiece` variable from your commented out example, since the temporary object definitely doesn't survive beyond that point. I am less sure if it can also be valid for the original single liner implementation, as I thought that the temporary objects are required to survive until the end of the full expression in the context of which they were created. But it can be a bug in the compiler (on top of the other bug that chooses the wrong overload).

mgautierfr: Interesting idea but not conclusive.
I have removed this constructor in the stringpiece header and same error occurs.

> I am less sure if it can also be valid for the original single liner implementation, as I thought that the temporary objects are required to survive until the end of the full expression in the context of which they were created. But it can be a bug in the compiler (on top of the other bug that chooses the wrong overload).

I confirm to you that single liner version is not helping.
A single line "fix" is `return icu::UnicodeString::fromUTF8(utf8EncodedString.data()).length();` but we may truncate binary data containing `\0` so I prefer be explicit about the data size.

mgautierfr:
Failing tests on Windows seems to be related to https://github.com/docopt/docopt.cpp/issues/49
The workaround is to tell docopt to use boost regex instead of std::regex.

But we don't have boost in kiwix-build...

mgautierfr:
kiwix-build now build docoptcpp with boost.regex on Windows.

src/zimcheck/zimcheck.cpp
```diff
@@ -43,36 +44,37 @@
...
+ -r --redundant       Redundant data check
+ -u --url_internal    URL check - Internal URLs
+ -x --url_external    URL check - External URLs
+ -d --details         Details of error
+ -b --progress        Print progress report
+ -j --json            Output in JSON format
+ -h --help            Displays Help
+ -v --version         Displays software version
+ -l --redirect_loop   Checks for the existence of redirect loops
+ -w=<nb_thread> --threads=<nb_thread>  count of threads to utilize [default: 1]
```
veloman-yunkan, line 70: Maybe it makes sense to mention in the commit message that short options are now accepted only in lower case (or make the short options UPPER CASE in the usage string, like before, and emphasize that they are accepted only as UPPER CASE).

src/zimcheck/zimcheck.cpp
```diff
@@ -82,11 +84,36 @@ std::string stringify(const T& x)
...
+    docopt::Options parsed_args;
+    try {
+      parsed_args = docopt::docopt_parse(
+        USAGE,
+        args_string,
+        false,
+        false);
+    } catch (docopt::DocoptArgumentError const& error) {
+		std::cerr << error.what() << std::endl;
+		std::cout << USAGE << std::endl;
```
veloman-yunkan, line 109: Shouldn't usage be printed t stderr too?

src/zimcheck/zimcheck.cpp
```diff
@@ -99,140 +126,69 @@ int zimcheck (const std::vector<const char*>& args)
...
     }
-
-    //Displaying Help for --help argument
-    if(help)
-    {
-        displayHelp();
+    
+    if (filename.empty()) {
+        std::cerr << "No file provided as argument" << std::endl;
+        std::cout << USAGE << std::endl;
```
veloman-yunkan, line 191: Same here

src/zimcheck/zimcheck.cpp
mgautierfr: There is `We lost parsing case insensitive options on the way.`.
But I can be more explicit.

I wonder why we need upper case option ?
Almost all tools I know use lower case option and never uppercase.

src/zimcheck/zimcheck.cpp
mgautierfr: I took care to change the less the behavior.
Test are checking that usage is print on output. I have done the same way.

src/zimcheck/zimcheck.cpp
veloman-yunkan: > There is `We lost parsing case insensitive options on the way.`.
Oops, didn't notice it.

> 
> I wonder why we need upper case option ? Almost all tools I know use lower case option and never uppercase.

I agree. But the old usage string documented upper case short options. If zimcheck were a popular tool with usage in a lot of scripts this change won't be welcome.

kelson42:
@mgautierfr This PR seems to brake packaing for Ubunut jammy and focal because they are working on `main`, see https://github.com/openzim/zim-tools/actions/runs/10540978671

mgautierfr:
Last commit fix compilation with older docoptcpp version found in jammy and focal distribution.

src/zimcheck/zimcheck.cpp
mgautierfr: >  If zimcheck were a popular tool with usage in a lot of scripts this change won't be welcome.

I agree. @kelson42 @rgaudin is it ok for you ?

kelson42: May I ask the technical reason why we have such a move in this PR?

mgautierfr: Because docopt don't do case insensitive parsing.

If we want to do it, we would have to rewrite usage string as:
```
-c -C --checksum        Internal CheckSum Test
```
instead of 
```
-c --checksum        Internal CheckSum Test
```

(I haven't check that multiple shorts options for only one long option is working)

kelson42: So, if this is confirmed that lowercase args would be better from the user perspective and habits AND we have no reasonable technical alternative, then I guess we can move forward like this but we need to make a major release!

mgautierfr: And I have just check and it seens to NOT work.

veloman-yunkan: At least we can preserve support for upper case short options (because that's how they were documented) instead of switching to lower case options.

kelson42: Lets keep uppercase short option only then

mgautierfr: Done

veloman-yunkan (commented):
Let's squash the last commit ("Move back to upper case short options.") into "Port zimcheck to docoptcpp.") and the PR can be merged.

mgautierfr:
Done

veloman-yunkan (approved):

PR #435, opened 2024-12-05
Fix zimwriterfs usage sample command

benoit74:
Sample command output in zimwriterfs command has two issues:
- it is not working (anymore?) because it misses the required `--name` parameter
- it does not propose a ZIM filename compliant with our conventions

This PR fixes both issues.


kelson42:
@benoit74 Thank you. I will try to fix the CI first, but your PR looks good.

benoit74:
No hurry at all, just an easy fix for a problem I stumbled upon which would have been a shame not to share ^^

PR #446, opened 2025-01-13
Fix buffer overrun in link normalisation

### LINKED ISSUE #439: tools.cpp::normalize_link can read off the end of the buffer
https://github.com/openzim/zim-tools/blob/768cbe534156a44abe7678493c12e52b2a3194a3/src/tools.cpp#L450

A malformed URL with `%` in the last 3 characters will cause the `p` pointer to be incremented past the null terminator.

ThisIsFineTM:
This fixes the buffer overrun issue. I changed the raw `char*` to an iterator and put a couple of bounds checks. First commit updates the normalize_link test which demonstrates the issue with EXPECT. Second commit applies the fix and changes the added tests to be ASSERT.

Closes #439

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -404,23 +404,24 @@ std::string normalize_link(const std::string& input, const std::string& baseUrl)
     output.reserve(baseUrl.size() + input.size() + 1);
 
     bool check_rel = false;
-    const char* p = input.c_str();
+    auto p = input.cbegin();
     if ( *(p) == '/') {
       // This is an absolute url.
       p++;
     } else {
       //This is a relative url, use base url
       output = baseUrl;
-      if (!output.empty() && output.back() != '/')
+      if (!output.empty() && output.back() != '/') {
           output += '/';
+      }
       check_rel = true;
     }
 
     //URL Decoding.
-    while (*p)
+    while (p < input.cend())
     {
         if ( check_rel ) {
-            if (strncmp(p, "../", 3) == 0) {
+            if (std::strncmp(&*p, "../", 3) == 0) {
                 // We must go "up"
                 // Remove the '/' at the end of output.
                 output.resize(output.size()-1);
@@ -432,7 +433,7 @@ std::string normalize_link(const std::string& input, const std::string& baseUrl)
                 check_rel = false;
                 continue;
             }
-            if (strncmp(p, "./", 2) == 0) {
+            if (std::strncmp(&*p, "./", 2) == 0) {
                 // We must simply skip this part
                 // Simply move after the ".".
                 p += 2;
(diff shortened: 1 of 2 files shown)


src/tools.cpp
```diff
@@ -404,23 +404,24 @@ std::string normalize_link(const std::string& input, const std::string& baseUrl)
...
       check_rel = true;
     }
 
     //URL Decoding.
-    while (*p)
+    while (p < input.cend())
     {
         if ( check_rel ) {
-            if (strncmp(p, "../", 3) == 0) {
+            if (std::strncmp(p.base(), "../", 3) == 0) {
```
veloman-yunkan, line 424: 1. Is `p.base()` a valid expression according to the C++ standard? If yes, when was `std::string::const_iterator::base()` first standardized?   
2. Can't this result in an out-of-bounds access? `c_str()` returns a null-terminated C-style string (which would be handled by `strncmp()` in an expected way) whereas with iterators there is no such guarantee (opening the door to potentially reading garbage bytes past the end of the string data).

src/tools.cpp
```diff
@@ -432,7 +433,7 @@ std::string normalize_link(const std::string& input, const std::string& baseUrl)
                 check_rel = false;
                 continue;
             }
-            if (strncmp(p, "./", 2) == 0) {
+            if (std::strncmp(p.base(), "./", 2) == 0) {
```
veloman-yunkan, line 436: Same here

src/tools.cpp
ThisIsFineTM: 1. `p.base()` returns the iterator's underlying type (in this case, a `const char*`) has been supported since the C++98 days (GCC4, Clang3) https://godbolt.org/z/fz7GrMcvE , but apparently it wasn't formally standardized until C++23 and MSVC doesn't like it https://en.cppreference.com/w/cpp/iterator/basic_const_iterator/base . Good catch. I haven't worked with MSVC for ~20 years so `::base` was something I took for granted. I'll update with a standard compliant version (`&*p` will give the same result https://godbolt.org/z/bnGEnWGqf).

2. `std::string::c_str()` returns a non-owning `const char*` to the underlying data stored within the std::string, and that is a requirement that has been standardized since ISO C++11 (https://godbolt.org/z/1fvKz6n4E). 

https://en.cppreference.com/w/cpp/string/basic_string/c_str
> c_str() and [data()](https://en.cppreference.com/w/cpp/string/basic_string/data) perform the same function. (since C++11)
> (Also see Example code)

https://en.cppreference.com/w/cpp/string/basic_string/data
> Returns a pointer to the underlying array serving as character storage.

std::string::iterator satisfies https://en.cppreference.com/w/cpp/named_req/ContiguousIterator, which has the requirement:
> *(a + n) is equivalent to *([std::addressof](http://en.cppreference.com/w/cpp/memory/addressof)(*a) + n).

A `const char*` retrieved by `string::c_str`/`string::data` and an iterator over a string are required to point to the same memory. 

So, no, I don't believe that it can result in an out of bounds access, but I can always add an explicit bounds check if that would relieve your concerns.

My longer term goal is to introduce c++17's std::string_view, which has a bit simpler semantics. This PR was just the minimum amount of code changes to fix the out of bounds issue.

ThisIsFineTM: I applied the change away from ::base. Tests are passing and valgrind is happy with it.

veloman-yunkan (commented):
LGTM, but please squash the last two commits into the initial two so that the code change history is cleaner.

ThisIsFineTM:
@veloman-yunkan squash pushed :+1:

PR #450, opened 2025-01-30
449 fix off by one

### LINKED ISSUE #449: zimcheck complains about percent-encoded URLs
URLs like `path/%26` are valid, aren’t they? When creating ZIM files for the DWDS project, we have files named like `&` (a lonesome ampersand) or `'` (apostrophe), and the right way to link to them seems to be `<a href="path/%26">…</a>`, `<a href="path/%27">…</a>`.

Minimal test case:

```
% cat test.html
<a href="%26">&amp;</a>
<a href="%27">&apos;</a>

% echo foo > "&"

% echo foo > "'"
```

After creating a ZIM file `test.zim` with `zimwriterfs`, the check via `zimcheck test.zim` fails with:

```
  The following links:
- %26
- %27
() were not found in article test.html
```

ThisIsFineTM:
Fixed the off-by-one error I introduced in normalize_link with `%` parsing and added a couple test cases to verify. 

Closes #449

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -450,8 +450,10 @@ std::string normalize_link(const std::string& input, const std::string& baseUrl)
 
         if ( *p == '%')
         {
-            if( (p+3) >= input.cend()){
+            if( (p+2) >= input.cend()){
                 // if the %XX token would go off the end of the string, just break
+                // string::end() returns an iterator pointing to the null terminator
+                // of the underlying c-string (guarenteed as of C++11)
                 break;
             }
             // hhx only officially supports hex unsigned char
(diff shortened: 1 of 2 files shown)


veloman-yunkan (changes requested):
It took me a while to figure out what was wrong with the old code. I think that an explanatory comment (reminding about the terminating `'\0'` character) is due.

ThisIsFineTM:
@veloman-yunkan Comment added and squashed.

veloman-yunkan:
> @veloman-yunkan Comment added and squashed.

I don't see it in the latest version of the PR

veloman-yunkan:
Also please fit the title of the second commit within the limits of the GitHub commit description format

ThisIsFineTM:
Huh. That was weird.

@veloman-yunkan Should be fixed now. I made the additional change to the commit message and added a little visual in the commit message to make the indexing more clear.

https://github.com/openzim/zim-tools/commit/506c2b41636a60b0dfa88d2d8a05a96336cc59a9

veloman-yunkan (approved):

kelson42:
@ThisIsFineTM Thank you very mich!

PR #452, opened 2025-02-04
Do not copy entry in 'W' namespace when recreating zim file.

mgautierfr:
Entries in 'W' namespace are automatically created by libzim creator. This is especially important as we will "lost" the namespace and so we would add the entry as an user entry.


veloman-yunkan (approved):
Frankly, I don't remember what purpose the `W` namespace serves, however I can't raise any objections against its addition to the list of other internal namespaces.

PR #456, opened 2025-02-25
[RFC] 444 mime types own unit

ThisIsFineTM:
Why: Centralize mimetype definitions so we aren't using raw string literals everywhere for mimetype checks.

Changes:
* mime types moved from `src/zimwriterfs/tools` to `src/mimetypes`
* compile-time known mimetypes are `inline constexpr std::string_view`
  * `inline constexpr` (c++17) will guarantee the respective literals have the same address / are deduplicated across translation units at link time
  * `std::string_view` (c++17) is a lightweight `constexpr` wrapper of a `char*` and `size`, with support for `std::string` interop, but since its `constexpr`, you aren't constructing a bunch of `std::string` objects at runtime to hold copies of the string literal data
* `extMimeTypes` is now a function returning `const&` to the `static std::map` ("static initialization order fiasco")
* `extMimeTypes` map now has a custom comparator for key lookups that will convert inputs to lower case, and otherwise act the same as `std::lexicographical_compare`
  * This allows the extension keys to be deduplicated since now there is no need to have all upper and all lower case entries.
* `zimwriterfs/tools::getMimeTypeForFile`:
  * refactored to use `std::string_view` `extMimeTypes` map entries
  * refactored implementation to be const-correct and exception-free.

### DIFF
--- a/src/mimetypes.h
+++ b/src/mimetypes.h
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2025 Kiwix <kiwix.org>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#ifndef OPENZIM_TOOLS_MIMETYPES_H
+#define OPENZIM_TOOLS_MIMETYPES_H
+
+#include <map>
+#include <string_view>
+
+inline constexpr std::string_view mimeTextHtml{"text/html"};
+inline constexpr std::string_view mimeImagePng{"image/png"};
+inline constexpr std::string_view mimeImageTiff{"image/tiff"};
+inline constexpr std::string_view mimeImageJpeg{"image/jpeg"};
+inline constexpr std::string_view mimeImageGif{"image/gif"};
+inline constexpr std::string_view mimeImageSvgXml{"image/svg+xml"};
+inline constexpr std::string_view mimeTextPlain{"text/plain"};
+inline constexpr std::string_view mimeTextXml{"text/xml"};
+inline constexpr std::string_view mimeAppEpubZip{"application/epub+zip"};
+inline constexpr std::string_view mimeAppPdf{"application/pdf"};
+inline constexpr std::string_view mimeAudioOgg{"audio/ogg"};
+inline constexpr std::string_view mimeVideoOgg{"video/ogg"};
+inline constexpr std::string_view mimeAppJavascript{"application/javascript"};
+inline constexpr std::string_view mimeAppJson{"application/json"};
+inline constexpr std::string_view mimeTextCss{"text/css"};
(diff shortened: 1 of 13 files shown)


src/mimetypes.h
```diff
@@ -0,0 +1,141 @@
...
+inline constexpr std::string_view extWebm       { "webm"};
+inline constexpr std::string_view extWebp       { "webp"};
+inline constexpr std::string_view extMp4        { "mp4"};
+inline constexpr std::string_view extDoc        { "doc"};
+inline constexpr std::string_view extDocx       { "docx"};
+inline constexpr std::string_view extPpt        { "ppt"};
+inline constexpr std::string_view extOdt        { "odt"};
+inline constexpr std::string_view extOdp        { "odp"};
+inline constexpr std::string_view extZip        { "zip"};
+inline constexpr std::string_view extWasm       { "wasm"};
```
veloman-yunkan, line 98: Why are these needed? They are not used anywhere outside `create_extMimeTypes()`.

src/mimetypes.h
```diff
@@ -0,0 +1,141 @@
...
+inline constexpr std::string_view mimeVideoWebm{"video/webm"};
+inline constexpr std::string_view mimeImageWebp{"image/webp"};
+inline constexpr std::string_view mimeVideoMp4{"video/mp4"};
+inline constexpr std::string_view mimeAppMSWord{"application/msword"};
+inline constexpr std::string_view mimeAppVndXmlWord{"application/vnd.openxmlforMats-officedocument.wordprocessingml.document"};
+inline constexpr std::string_view mimeAppVndMSPpt{"application/vnd.ms-powerpoint"};
+inline constexpr std::string_view mimeVndOODoc{"application/vnd.oasis.openDocument.text"};
+inline constexpr std::string_view mimeAppZip{"application/zip"};
+inline constexpr std::string_view mimeAppWasm{"application/wasm"};
+inline constexpr std::string_view mimeAppOctetStream{"application/octet-stream"};
```
veloman-yunkan, line 58: Most of these are not used outside `create_extMimeTypes()` and therefore their existence is not justified.

veloman-yunkan (changes requested):
The second commit is too large for me. I didn't review it meticulously. Please implement the PR as a series of small self-contained changes. I see the following steps:

1. Address the "static initialization order fiasco"
2. Change the type of `extMimeTypes()` (introduce the case-insensitive custom comparator) and get rid of the UPPERCASE entries in `_create_extMimeTypes()`
3. Rewrite `getMimeTypeForFile()` without abusing `try/catch` statements
4. Introduce mimetypes.h/cpp and move relevant code there

src/mimetypes.h
ThisIsFineTM: For this in particular, I felt that there should be a standardized way to refer to the individual mimetypes in the code, and only having the minimal viable subset of mimetypes defined felt somewhat half-baked. I was looking at the other openzim projects like libzim/xapian, and there are a lot more uses of specific mimetypes, so I wanted the definitions to be easily translatable if you decide its worth the time to upstream this in the dependency hierarchy while also supporting future development.

This is also why I marked this PR as RFC. I wanted to get your take on it in a place I wouldn't be causing too much trouble first :)

I'm fine with just changing out the extMimeType map comparator to remove the duplicate entries and calling it a day if you think the individual symbol definitions is going overboard.

veloman-yunkan: > I'm fine with just changing out the extMimeType map comparator to remove the duplicate entries and calling it a day if you think the individual symbol definitions is going overboard.

Let's proceed that way (the title and description of the PR will need to be updated).

ThisIsFineTM: Sounds good. I'll stash the other changes in a separate branch in my fork, apply the minimal changes, and update the issue/PR title to accurately reflect the work.

ThisIsFineTM: On second thought, I am going to just close this PR and put the case insensitive map comparator into its own PR. We had a good discussion on the merits of putting the mime types into their own unit and I'd like to keep that preserved with the relevant title in case we decide to revisit it later.

PR #458, opened 2025-03-11
Case-insensitive comparator for extMimeTypes.

### LINKED ISSUE #457: Case insensitive handling of file extensions during MIME-type detection
Currently file extensions resolve to registered MIME-types only if they are fully in lower case or fully in upper case. Why don't we simply make the handling of file extensions case insensitive?

ThisIsFineTM:
This introduces a case-insensitive lexicographical comparator for the mimetype map in zimwriterfs/tools. This allows us to remove the duplicate entries in the mime type map. 

Closes #457.

### DIFF
--- a/src/zimwriterfs/tools.cpp
+++ b/src/zimwriterfs/tools.cpp
@@ -26,91 +26,82 @@
 #include <iostream>
 #include <iomanip>
 #include <map>
+#include <string_view>
 
 #include <zlib.h>
 #include <magic.h>
 
-/* Init file extensions hash */
-static std::map<std::string, std::string> _create_extMimeTypes()
-{
-  std::map<std::string, std::string> extMimeTypes;
-  extMimeTypes["HTML"] = "text/html";
-  extMimeTypes["html"] = "text/html";
-  extMimeTypes["HTM"] = "text/html";
-  extMimeTypes["htm"] = "text/html";
-  extMimeTypes["PNG"] = "image/png";
-  extMimeTypes["png"] = "image/png";
-  extMimeTypes["TIFF"] = "image/tiff";
-  extMimeTypes["tiff"] = "image/tiff";
-  extMimeTypes["TIF"] = "image/tiff";
-  extMimeTypes["tif"] = "image/tiff";
-  extMimeTypes["JPEG"] = "image/jpeg";
-  extMimeTypes["jpeg"] = "image/jpeg";
-  extMimeTypes["JPG"] = "image/jpeg";
-  extMimeTypes["jpg"] = "image/jpeg";
-  extMimeTypes["GIF"] = "image/gif";
-  extMimeTypes["gif"] = "image/gif";
-  extMimeTypes["SVG"] = "image/svg+xml";
-  extMimeTypes["svg"] = "image/svg+xml";
-  extMimeTypes["TXT"] = "text/plain";
-  extMimeTypes["txt"] = "text/plain";
-  extMimeTypes["XML"] = "text/xml";
-  extMimeTypes["xml"] = "text/xml";
-  extMimeTypes["EPUB"] = "application/epub+zip";
-  extMimeTypes["epub"] = "application/epub+zip";
-  extMimeTypes["PDF"] = "application/pdf";
-  extMimeTypes["pdf"] = "application/pdf";
-  extMimeTypes["OGG"] = "audio/ogg";
-  extMimeTypes["ogg"] = "audio/ogg";
-  extMimeTypes["OGV"] = "video/ogg";
-  extMimeTypes["ogv"] = "video/ogg";
-  extMimeTypes["JS"] = "application/javascript";
-  extMimeTypes["js"] = "application/javascript";
-  extMimeTypes["JSON"] = "application/json";
-  extMimeTypes["json"] = "application/json";
-  extMimeTypes["CSS"] = "text/css";
-  extMimeTypes["css"] = "text/css";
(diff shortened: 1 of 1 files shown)


src/zimwriterfs/tools.cpp
```diff
@@ -30,87 +30,85 @@
...
-  extMimeTypes["ODP"] = "application/vnd.oasis.opendocument.text";
-  extMimeTypes["zip"] = "application/zip";
-  extMimeTypes["ZIP"] = "application/zip";
-  extMimeTypes["wasm"] = "application/wasm";
-  extMimeTypes["WASM"] = "application/wasm";
-
-  return extMimeTypes;
+  return MimeMap_t{
+      {"html", "text/html"},
+      {"htm", "text/html"},
```
veloman-yunkan, line 73: By taking advantage of c++11 style initialization, the `_create_extMimeTypes()` function is rendered meaningless - you can initialize `extMimeTypes` directly.

src/zimwriterfs/tools.cpp
```diff
@@ -30,87 +30,85 @@
...
+      {"ppt", "application/vnd.ms-powerpoint"},
+      {"odt", "application/vnd.oasis.opendocument.text"},
+      {"odp", "application/vnd.oasis.opendocument.text"},
+      {"zip", "application/zip"},
+      {"wasm", "application/wasm"},
+  };
 }
 
-static std::map<std::string, std::string> extMimeTypes = _create_extMimeTypes();
+static MimeMap_t extMimeTypes = _create_extMimeTypes();
```
veloman-yunkan, line 111: Let's make this `const`.

src/zimwriterfs/tools.cpp
```diff
@@ -30,87 +30,85 @@
...
+            if (charCompare(*first2, *first1)) {
+                return false;
+            }
+        }
+
+        return (first1 == last1) && (first2 != last2);
+    }
+};
+
+using MimeMap_t = std::map<std::string, const std::string, tolower_str_comparator>;
```
veloman-yunkan, line 66: This seems too much code for the problem addressed. With the current limited usage of `extMimeTypes` and some rewrite you can achieve the same effect as follows:

```c++
static const std::string& getMimeType(std::string ext) {
    static const std::map<std::string, const std::string> extMimeTypes = {
        {"html", "text/html"},
        ...
    };

    for ( size_t i = 0; i < ext.size(); ++i )
        ext[i] = std::tolower(ext[i]);

    return extMimeTypes.at(ext);
}
```

veloman-yunkan (changes requested):

src/zimwriterfs/tools.cpp
ThisIsFineTM: I'll move that initialization to be direct and force push over the previous commit.

src/zimwriterfs/tools.cpp
ThisIsFineTM: There are more places where the custom comparator provides enhancements, most notably usage of `std::map::find`. I'd like to move away from relying on exceptions in cases where `find` can get the job done. I'm just breaking up the work into smaller PRs so the individual changes are less time consuming to review. I could provide helpers to the mimetype map to also work with find, but at that point we're hitting the same amount of lines as the comparator with the added detriment of having to make a copy of the `ext` string rather than getting it in-place for each query.

veloman-yunkan: The problem with custom comparator is that it results in a new instantiation of the std::map template, bloating the binary code.

> I'd like to move away from relying on exceptions in cases where `find` can get the job done.

I too frown upon such usage of exceptions, and in this case it can be worked around by making `getMimeType()` return an empty string (which will have to be checked by the caller).

> with the added detriment of having to make a copy of the ... `ext` string rather than getting it in-place for each query.

I believe you know the adage about premature optimization. The problem with it is that when tiny optimizations are achieved at the cost of added complexity the final result may be worse performance compared to a simpler solution (or a relatively easy modification of it that addresses bigger performance problems).

ThisIsFineTM: As a side note, if binary bloat is more of a concern, enabling LTO can cut the tool binary sizes by roughly 10-50% depending on which compiler is used (tested with gcc 14.2 and clang 19.1). Just need to put in a 1 line fix into the metadata test file to make the linker happy. I can do that in another PR after we resolve this one.

veloman-yunkan: @ThisIsFineTM I appreciate all the effort that you put into defending your solution. My point of view is that one should stick to the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) unless there are good reasons to deviate from it. My proposal to encapsulate `extMimeMap` behind a single `getMimeType()` API (with `getMimeType(ext).empty()` serving as a test for `ext` not being a known extension) is along those lines.

However I don't mind if your solution is merged upon someone else's approval - IMHO the issue doesn't deserve that much discussion.

ThisIsFineTM: @veloman-yunkan Yeah, we are bikeshedding a bit. Although the find with how much LTO can reduce the tool size was a nice perk. I'll open up a separate issue for that and put in a PR.

@kelson42 @mgautierfr I know you two are busy, but if one of you could take a glance at this at some point and give a cursory :+1: or :-1: that would be appreciated.

veloman-yunkan:
> @veloman-yunkan I believe we should do a new review pass

@kelson42 Nothing has changed since the last review (weather & geopolitical events are irrelevant to this PR).

Issue #305, opened 2022-06-20
Core dump checking latest stackoverflow


rgaudin:
`time zimcheck -A /data/download/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim`

```
[INFO] Checking zim file /data/download/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim
[INFO] Zimcheck version is 3.1.1
[INFO] Verifying ZIM-archive structure integrity...
[INFO] Checking zim file /data/download/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim
[INFO] Zimcheck version is 3.1.1
[INFO] Verifying ZIM-archive structure integrity...
[INFO] Avoiding redundant checksum test (already performed by the integrity check).
[INFO] Searching for metadata entries...
[INFO] Searching for Favicon...
[INFO] Searching for main page...
[INFO] Verifying Articles' content...
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_replace_aux
Aborted (core dumped)

real    35m43.516s
user    31m19.384s
sys     5m24.876s
```

```
zimcheck --version
zim-tools 3.1.1

libzim 7.2.2
+ libzstd 1.5.2
+ liblzma 5.2.4
+ libxapian 1.4.18
+ libicu 58.2.0
```

rgaudin:
@kelson42 no I don't but it looks like the log is giving a hint about what's happening and where. Hopefully that's enough…

veloman-yunkan:
> @kelson42 no I don't but it looks like the log is giving a hint about what's happening and where. Hopefully that's enough…

Unfortunately, the information in the log only tells about an out-of-bounds access during a mutating operation on a string object, which is absolutely not enough to pinpoint the root cause.

@kelson42 Can I debug this remotely?

veloman-yunkan:
@kelson42 I feel that you missed my feedback. Is there a way for me to debug this crash remotely (without having to download the ZIM)?

kelson42:
@veloman-yunkan I don't have forgotten you. But the we have neither a reproduction case with a smaller ZIM file nor a dedicated online env. we could give you access to.

Usually and so far, most of us (devs) have access to private online linux servers we do use as playground or debug environnements. This kind of need has been solved that way.

[...]

PR #488, opened 2026-01-11
zimcheck: fix isOutofBounds bug related to URL queries

IMayBeABitShy:
Greetings,

it's been a while since I've done anything with C++, so feel free to tear the suggested changes of this PR apart.
This is just a simple "I've found a bug and fixed it" PR otherwise.

# Bug report (fix follows in the next section)

**Observed Behavior:**

I've recently found `zimcheck` failing with the error message indicating that an URL was out of bounds despite being valid. 
Let's assume the following ZIM structure:

```
ZIM_ROOT (/)
 - tools/
    - tool.html
 - dir/
    - subdir/
      - index.html
      - file.html
```

The URL was something along the lines of `../../tools/tool.html?path=../dir/subdir/file.html` with the base directory being `/dir/subdir/index.html`. Now, this URL should resolve to `/tools/tool.html?path=../dir/subdir/file.html`. Entirely valid, yet `zimcheck` fails due to wrongly identifying the above link as out of bounds (meaining that it reaches beyond the ZIM root directory).

Here's a minimal example of the failing code:

```c++
int main() {
    std::string inp = "../../tool.html?path=../dir/subdir/file.html";
    std::string path = "dir/subdir/index.html";
    auto baseUrl = path;
    auto pos = baseUrl.find_last_of('/');
    baseUrl.resize( pos==baseUrl.npos ? 0 : pos );

    cout << "path: " << path << "\n";
    cout << "base: " << baseUrl << "\n";
    cout << "inp:  " << inp << "\n";

    if (isOutofBounds(inp, baseUrl)) {
        cout << "OOB: true\n";
    } else {
        cout << "OOB: false\n";
    }
    return 0;
}
```

**Expected behavior:**

As you can probably guess from the previous subsection, the expected behavior is that the link shouldn't be identified as an out of bounds link. Consequently, `zimcheck` should not fail the zim.

# The fix

The problem lies in the `isOutofBounds` function, which doesn't properly parse the HTML links. Rather, it interprets the links as filesystem paths. Which is probably sort-of valid for some file types, but for HTML files the links must be interpreted as hyperlinks. The `isOutofBounds` function is rather crude in that it only compares to number of occourences of `/` and ` ../` against each other rather than actualyl understanding how a hyperlink URL may be structured.

~The bug in this PR can be fixed by limiting the range in which the `../` occurences are counted by the first occurence of `?`.~ The bug in this PR can be fixed by properly extracting the path component of the URL and only checking isOutOfBounds on that part.

~**Important note::** this fix is just as "crude" as the original function. It doesn't really resolve the underlying issue - that being the lack of understanding of how a URL works - and thus doesn't fix like a dozen other potential issues I can think of. For example, should a URL specify a fragment containing `../`, this same issue may occur.~

# Other stuff

This PR contains a new function `extractPathFromLink`, which is used to get the path component of a http URL, as well as tests for both this new function and the bug described in this issue.

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -377,7 +377,8 @@ bool isOutofBounds(const std::string& input, std::string base)
     //count nr of substrings ../
     int nrsteps = 0;
     std::string::size_type pos = 0;
-    while((pos = input.find("../", pos)) != std::string::npos) {
+    std::string::size_type queryPos = input.find("?");
+    while((pos = input.find("../", pos)) != std::string::npos && (pos <= queryPos || queryPos == std::string::npos)) {
         nrsteps++;
         pos += 3;
     }
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -363,12 +363,47 @@ std::vector<html_link> generic_getLinks(const std::string& page)
     return links;
 }
 
+std::string extractPathFromLink(const std::string& link)
+{
+    // Reminder: URL format for hyperrefs is: scheme://user:password@host:port/path?query#fragment
+    // Note: this function assumes that the provided URL is a hyperref and should only be applied on such URLs
+    // strip scheme
+    auto schemeI = link.find("://");
+    if (schemeI != std::string::npos) {
+        schemeI = schemeI + 3;
+    } else {
+        schemeI = 0;
+    }
+    // strip user+password
+    auto authorityEndI = link.find_first_of("/?#", schemeI);
+    auto passwordI = link.find("@", schemeI);
+    if (passwordI != std::string::npos && passwordI < authorityEndI) {
+        passwordI = passwordI + 1;
+    } else {
+        passwordI = schemeI;
+    }
+    // find path
+    auto firstSlashI = link.find("/", passwordI);
+    auto pathEndI = link.find_first_of("?#", passwordI);
+    // a path must start with a / if there is a :// in the URL
+    // if there is no ://, then the path, a path may start without a leading /, but does so immediately
+    std::string path;
+    if (schemeI == 0) {
+        path = link.substr(0, pathEndI);
+    } else {
+        path = link.substr(firstSlashI, pathEndI-firstSlashI);
+    }
+    return path;
+}
+
 bool isOutofBounds(const std::string& input, std::string base)
 {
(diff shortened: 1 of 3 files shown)


kelson42:
@IMayBeABitShy @veloman-yunkan What is the status here? I had the feeling we are almost at the end of the effort!?

IMayBeABitShy:
@kelson42 Well, the bug I've found is fixed with the current PR, so I considered the code part done. However, veloman-yunkan stated:

> I guess the possibility of links containing search and/or fragment components has been completely overlooked. We now have to define how to handle them. My feeling is that in the context of zimcheck they should be simply discarded.

So there was still some clarity needed how such links should be handled. I've written my opinion/view of this in the previous comment. However, perhaps we should consider adding validation for fragments, too. E.g. fail the zimcheck if a link like `foo/bar.html#baz` was found but `bar.html` does not contain any element with id `baz`.

benoit74:
@IMayBeABitShy PoV expressed in https://github.com/openzim/zim-tools/pull/488#issuecomment-3769369311 (and previous comments) is aligned with zimit / warc2zim and other scrapers behaviors, and what I consider is the right approach. It is scraper responsibility to URL-encode in HTML, otherwise we need to consider we are dealing with just a regular query parameter or fragment.

IMayBeABitShy:
Well, as far as I can tell, there's nothing preventing this PR from being merged.

veloman-yunkan:
> Well, as far as I can tell, there's nothing preventing this PR from being merged.

@kelson42 I have to revisit the PR and the discussion to be able to tell if I agree with @IMayBeABitShy. Please give me a couple of days.

veloman-yunkan:
> BTW, I remember that there once has been a similar discussion of relative links going out of bounds (ascending with a series of `../` above the top location) and it was pointed out that it is a perfectly valid URL according to the spec. I will try to find that comment exchange

FYI, I found the mentioned discussion; it happens around https://github.com/openzim/zim-tools/issues/305#issuecomment-1172638533

veloman-yunkan:
@IMayBeABitShy FYI, #511 is going to override this PR. You are welcome to review it.

kelson42:
Superseeded by #511

PR #499, opened 2026-03-23
fix: modify getTextLength to use BreakIterator for grapheme count

### LINKED ISSUE #498: Count graphemes, not unicode code points
Reported by @Pranavjeet-Naidu in https://github.com/openzim/zim-tools/pull/497#issuecomment-4107836762

Can you explain where this function is used? I feel like you are probably right.

Pranavjeet-Naidu:
Fixes #498

Used ICU's BreakIterator, took the definition code from their documentation + added a status check as well. 
Please suggest changes if needed, considering this is my first time working with this.

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -33,6 +33,8 @@
 #include <algorithm>
 #include <regex>
 #include <array>
+#include <unicode/brkiter.h>
+#include <unicode/utypes.h>
 #include <unicode/unistr.h>
 
 #ifdef _WIN32
@@ -680,6 +682,23 @@ size_t getTextLength(const std::string& utf8EncodedString)
   // This is broken
   // icu::StringPiece stringPiece(utf8EncodedString);
   // This is not
-  icu::StringPiece stringPiece(utf8EncodedString.data(), static_cast<int32_t>(utf8EncodedString.size()));
-  return icu::UnicodeString::fromUTF8(stringPiece).length();
+  icu::StringPiece stringPiece(utf8EncodedString.data(),
+                               static_cast<int32_t>(utf8EncodedString.size()));
+  icu::UnicodeString ustr = icu::UnicodeString::fromUTF8(stringPiece);
+
+  UErrorCode status = U_ZERO_ERROR;
+  std::unique_ptr<icu::BreakIterator> bi(
+      icu::BreakIterator::createCharacterInstance(icu::Locale::getRoot(), status));
+
+  if (!U_SUCCESS(status)) {
+    return ustr.length();  // Fallback to codepoint count
+  }
+
+  bi->setText(ustr);
+
+  size_t count = 0;
+  while (bi->next() != icu::BreakIterator::DONE) {
+    ++count;
+  }
+  return count;
 }
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <regex>
 #include <array>
+#include <unicode/unistr.h>
 
 #ifdef _WIN32
 #define SEPARATOR "\\"
@@ -662,4 +663,23 @@ std::string httpRedirectHtml(const std::string& redirectUrl)
 bool guess_is_front_article(const std::string& mimetype) {
   return ( mimetype.find("text/html") == 0
         && mimetype.find("raw=true") == std::string::npos);
+}
+
+size_t getTextLength(const std::string& utf8EncodedString)
+{
+  // For some unknown reason implicite convertion from std::string to icu::StringPiece
+  // is broken on Windows.
+  // Constructors are definde in stringpiece.h as
+  // ```
+  // StringPiece(const std::string& str)
+  //  : ptr_(str.data()), length_(static_cast<int32_t>(str.size())) { }
+  // StringPiece(const char* offset, int32_t len) : ptr_(offset), length_(len) { }
+  // ```
+  // However using the first constructor ends with a corrupted StringPiece (wrong ptr)
+  // and using second one works. Don't ask me why
+  // This is broken
+  // icu::StringPiece stringPiece(utf8EncodedString);
+  // This is not
+  icu::StringPiece stringPiece(utf8EncodedString.data(), static_cast<int32_t>(utf8EncodedString.size()));
+  return icu::UnicodeString::fromUTF8(stringPiece).length();
 }
--- a/test/meson.build
+++ b/test/meson.build
@@ -1,7 +1,7 @@
 gtest_dep = dependency('gtest', main:true, fallback:['gtest', 'gtest_main_dep'], required:false)
 
 
-test_deps = [gtest_dep, libzim_dep, icu_dep, docopt_dep]
+test_deps = [gtest_dep, libzim_dep, icu_uc_dep, icu_dep, docopt_dep]
 tests = [
     'metadata-test',
     'zimcheck-test'
--- a/test/meson.build
+++ b/test/meson.build
@@ -21,7 +21,7 @@ zimwriter_srcs = [  '../src/zimwriterfs/tools.cpp',
 
 tests_src_map = { 'zimcheck-test' : ['../src/zimcheck/zimcheck.cpp', '../src/zimcheck/checks.cpp',  '../src/zimcheck/json_tools.cpp', '../src/tools.cpp', '../src/metadata.cpp'],
                   'tools-test' : zimwriter_srcs,
-                  'metadata-test' : ['../src/metadata.cpp'],
+                  'metadata-test' : ['../src/metadata.cpp', '../src/tools.cpp'],
                   'zimwriterfs-zimcreatorfs' : zimwriter_srcs }
 
 if gtest_dep.found() and not meson.is_cross_build()
--- a/meson.build
+++ b/meson.build
@@ -25,6 +25,7 @@ with_xapian_support = compiler.has_header_symbol(
 find_library_in_compiler = meson.version().version_compare('>=0.31.0')
 rt_dep = dependency('rt', required:false)
 docopt_dep = dependency('docopt', static:static_linkage)
+icu_uc_dep = dependency('icu-uc', static:static_linkage)
 icu_dep = dependency('icu-i18n', static:static_linkage)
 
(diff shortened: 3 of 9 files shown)


veloman-yunkan (changes requested):
This fix begs a unit test. Therefore the PR should be implemented as follows:

Commit 1: move the function `getTextLength()` to `src/tools.cpp` and declare it in `src/tools.h`.
Commit 2: add a unit test in `test/tools-test.cpp`. It must include a test-point reflecting the bug being addressed.
Commit 3: fix the bug

src/tools.cpp
```diff
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <regex>
 #include <array>
+#include <unicode/utypes.h>
```
veloman-yunkan, line 37: Why is this needed in this commit?

veloman-yunkan (changes requested):
> @veloman-yunkan, the three commits like you've asked are up, once you give a go ahead, I will squash them as well.

The commits must not be squashed. The PR must consist of the said three commits at the time of merge. Now there are four commits.

Please also don't abuse the "fix:" label in the commit title. The commit message must not be misleading - adding a unit test is not a fix. Also it must be informative. For example, if a new unit test fails because it demonstrates a pre-existing bug, that must be stated in the commit description.

src/tools.cpp
Pranavjeet-Naidu: I've moved it away into the 3rd commit.

Pranavjeet-Naidu:
@veloman-yunkan I don't understand why the tests are failing in the CI, I've run them myself locally: 

```
grass@pranavjeets-MacBook-Air ~/p/zim-tools (fix/correct_grapheme_count)> meson test -C build 
ninja: Entering directory `/Users/grass/projects/zim-tools/build'
ninja: no work to do.
1/4 metadata-test                   OK              0.02s
2/4 tools-test                      OK              0.02s
3/4 zimwriterfs-zimcreatorfs        OK              0.09s
4/4 zimcheck-test                   OK              0.64s

Ok:                4   
Fail:              0   

Full log written to /Users/grass/projects/zim-tools/build/meson-logs/testlog.txt
```

while the Windows CI fail as well, googling around, the only possible reason I see is icu-uc not being pre-installed, please confirm this, could use some help here.

I went ahead and checked: 
uses: kiwix/kiwix-build/actions/dl_deps_archive@... ( from .github/ci.yml)
then, in kiwix-build/kiwixbuild/dependencies/zim-tools.py :

```
from .base import Dependency, GitClone, MesonBuilder
from kiwixbuild._global import neutralEnv


class ZimTools(Dependency):
    name = "zim-tools"
    force_build = True

    class Source(GitClone):
        git_remote = "https://github.com/openzim/zim-tools.git"
        git_dir = "zim-tools"

    class Builder(MesonBuilder):
        @classmethod
        def get_dependencies(cls, configInfo, allDeps):
            base_deps = ["libzim", "docoptcpp", "mustache"]
            if neutralEnv("distname") != "Windows":
                base_deps += ["libmagic", "gumbo"]
            return base_deps

        @property
        def configure_options(self):
            # We don't build zimwriterfs on Windows, and so we don't have magic
            if neutralEnv("distname") != "Windows":
                yield f"-Dmagic-install-prefix={self.buildEnv.install_dir}"
            if self.buildEnv.configInfo.static:
                yield "-Dstatic-linkage=true"
```

wouldn't this mean that the CI will fail until the icu headers are added.

veloman-yunkan:
> the only possible reason I see is icu-uc not being pre-installed

@Pranavjeet-Naidu Welcome to the intricate world of our build infrastructure!

`icu-uc` is included in https://tmp.kiwix.org/ci/deps_linux-x86_64-static_zim-tools.tar.gz which is used by the failed **[CI / Linux (linux-x86_64-static)](https://github.com/openzim/zim-tools/actions/runs/23815280858/job/69413019111?pr=499)** job and contains `BUILD_linux-x86_64/INSTALL/lib/x86_64-linux-gnu/libicuuc.a`.

However we build ICU (as well as other dependencies) ourselves and omit most of its data to minimize the size of some of our build artifacts for downstream projects. See kiwix/kiwix-build#494 which had to deal with a similar problem.

@kelson42 Here we have a situation where additional data must be included in the ICU library for this PR to work. How should we proceed?

Pranavjeet-Naidu:
> > the only possible reason I see is icu-uc not being pre-installed
> 
> @Pranavjeet-Naidu Welcome to the intricate world of our build infrastructure!
> 
> `icu-uc` is included in https://tmp.kiwix.org/ci/deps_linux-x86_64-static_zim-tools.tar.gz which is used by the failed **[CI / Linux (linux-x86_64-static)](https://github.com/openzim/zim-tools/actions/runs/23815280858/job/69413019111?pr=499)** job and contains `BUILD_linux-x86_64/INSTALL/lib/x86_64-linux-gnu/libicuuc.a`.
> 
> However we build ICU (as well as other dependencies) ourselves and omit most of its data to minimize the size of some of our build artifacts for downstream projects. See [kiwix/kiwix-build#494](https://github.com/kiwix/kiwix-build/pull/494) which had to deal with a similar problem.
> 
> @kelson42 Here we have a situation where additional data must be included in the ICU library for this PR to work. How should we proceed?

Please correct me if I'm wrong but here, the fix was to rebuild the entire dependency tree. 
The other way is to maybe have icu in a meson wrap but then we would have to compile and install it for every CI run involving this code. Can we skip the compilation part and modify the CI runner accordingly then?

Pranavjeet-Naidu:
@kelson42 any updates on how to proceed?

veloman-yunkan:
> Please correct me if I'm wrong but here, the fix was to rebuild the entire dependency tree. The other way is to maybe have icu in a meson wrap but then we would have to compile and install it for every CI run involving this code. Can we skip the compilation part and modify the CI runner accordingly then?

@Pranavjeet-Naidu ~We will just need to update https://github.com/kiwix/kiwix-build/tree/main/kiwixbuild/patches/icu4c_custom_data.patch once we make the decision to bundle more data with our binaries.~

Updated 2026-08-05: It turns out that was on old patch that is no longer used. The data must now be managed somewhat differently.

kelson42:
@Pranavjeet-Naidu Thank you very much for your work and your patience. This is really important. Sorry for been a bit unreactive, it is just a lack of time. I will try to do better from now on zim-tools and your other PRs.

@veloman-yunkan Regarding data mgmt in libicu, here is how it should work (not sure how it works effectively):
* We create a subset of the data needed by libicu, with only what we need
* For multiarch binaries, at least Android, we don't compile `libicu` with the data embedded. We deliver the data file separately and it is then loaded separately at runtime (this allows to have only once the dataset in/beside the final binary)
* Otherwise, for all other cases, the data is just embedded in libicu
* icu-c problematic is AFAIK totally unrelated to the data topic, it is just that latest version of libicu have been structured sigltly differently and now require to be included explicitly.
 
Here not sure what goes wrong, but might be (at least) https://github.com/kiwix/kiwix-build/issues/750. But for sure the data should be embedded in all libicu builds for zim-tools (because none of zim-tools builds are multiarch AFAIK). Would that be possible to fix that, to enable the merge of this important PR?

kelson42:
FYI, we are about to start implementation of https://github.com/openzim/cms/issues/221 and hopefully we can have this merged before it is implemented.

kelson42:
> `icu-uc` is included in https://tmp.kiwix.org/ci/deps_linux-x86_64-static_zim-tools.tar.gz which is used by the failed **[CI / Linux (linux-x86_64-static)](https://github.com/openzim/zim-tools/actions/runs/23815280858/job/69413019111?pr=499)** job and contains `BUILD_linux-x86_64/INSTALL/lib/x86_64-linux-gnu/libicuuc.a`.
> 
> However we build ICU (as well as other dependencies) ourselves and omit most of its data to minimize the size of some of our build artifacts for downstream projects. See [kiwix/kiwix-build#494](https://github.com/kiwix/kiwix-build/pull/494) which had to deal with a similar problem.
> 
> @kelson42 Here we have a situation where additional data must be included in the ICU library for this PR to work. How should we proceed?

@veloman-yunkan We have here no pressure to save storage space (this is not running on a mobile device), the libicu data should be full and included within the binary. I wonder why this is not the case.

kelson42:
@veloman-yunkan @Pranavjeet-Naidu Not sure about the next step here. Coukd we please be explicit about it?

Pranavjeet-Naidu:
> @veloman-yunkan @Pranavjeet-Naidu Not sure about the next step here. Coukd we please be explicit about it?

> @veloman-yunkan We have here no pressure to save storage space (this is not running on a mobile device), the libicu data should be full and included within the binary. I wonder why this is not the case.

I agree with this as well, I would prefer if we just include the entirety of libicu given there are no particular constraints that I can think of.

veloman-yunkan:
kiwix/kiwix-build#959 seems to have fixed (at least for the Linux build) the issue with ICU data - see https://github.com/openzim/zim-tools/actions/runs/31087559120 which has used the artifacts of kiwix/kiwix-build#959 vs https://github.com/openzim/zim-tools/actions/runs/29150871992. I will now investigate what's wrong with the Windows build.

veloman-yunkan:
This issue with the Windows build has been addressed in kiwix/kiwix-build#959 as well. Once that PR is merged this PR must be rebased on main with the temporary commit removed.

test/meson.build
```diff
@@ -21,7 +21,7 @@ zimwriter_srcs = [  '../src/zimwriterfs/tools.cpp',
 
 tests_src_map = { 'zimcheck-test' : ['../src/zimcheck/zimcheck.cpp', '../src/zimcheck/checks.cpp',  '../src/zimcheck/json_tools.cpp', '../src/tools.cpp', '../src/metadata.cpp'],
                   'tools-test' : zimwriter_srcs,
-                  'metadata-test' : ['../src/metadata.cpp'],
+                  'metadata-test' : ['../src/metadata.cpp', '../src/tools.cpp'],
```
veloman-yunkan, line 24: Shouldn't this change belong to the first commit (where `getTextLength()` is moved)?

veloman-yunkan (approved):

Issue #464, opened 2025-04-16
zimcheck: report all links in error in HTML files


benoit74:
While checking internal and external links in HTML items (at least), zimcheck reports only one error per ZIM entry: https://github.com/openzim/zim-tools/blob/3ad99a1682a2da084d5d6e86bc1f8986a7b149c6/src/zimcheck/checks.cpp#L446 and https://github.com/openzim/zim-tools/blob/3ad99a1682a2da084d5d6e86bc1f8986a7b149c6/src/zimcheck/checks.cpp#L459

While this probably makes sense from a perspective where one wants no error at all, it also make it more complex to have an oversight of all issues in a given ZIM (we have to fix first error, run zimcheck again, potentially multiple time until we've exhausted all errors). 

It also makes it hard to cope with "acceptable" errors which might hide real issue. This is my case with a zimit ZIM where the error reported on all items is an error about a missing JS resources which has been blocked by browsertrix addblocker. This first error is "acceptable" in the sense that ZIM still works pretty well from a user perspective. However on the specific ZIM I'm diagnosing, there are few (or many, I don't know) other errors which are "not acceptable": missing HTML pages making navigation broken. I can't use zimcheck to find these errors.

I would like to have an option like "report all" which allows to report all errors on all HTML items in a given ZIM.

Pranavjeet-Naidu:
@benoit74 Given that external errors are relatively insignificant compared to internal links errors, having something like
 ```--report all ``` would mean that we have in situations when I only need internal errors, I get external reporting as well. Do you think it would be better to have 2 different flags : ``` --report-all-internal ``` and ``` --report-all-external```?

The code already supports this btw, atleast on the actual reporter side: 
```cpp
void ErrorLogger::addMsg(MsgId msgid, const MsgParams& msgParams)
{
  const MsgInfo& m = msgTable.at(msgid);
  setTestResult(m.check, false);
  reportMsgs[size_t(m.check)].push_back({msgid, msgParams});
}
```
and then looking at msgTable's definition:, these mappings already exist: 
URL_INTERNAL -> DANGLING_LINKS, OUTOFBOUNDS_LINK
URL_EXTERNAL -> EXTERNAL_LINK

benoit74:
I see advantages to both approaches, not sure how to decide on this. 

@kelson42 @veloman-yunkan do you have inputs on this question?

Pranavjeet-Naidu:
[...]

PR #500, opened 2026-04-20
Enhancement: memsafe zimcheck reporting

### LINKED ISSUE #311: Make zimcheck report its findings without accumulating them all in memory
zimcheck's traditional (non-JSON) report format enforces an implementation where all findings (errors/warnings) are accumulated in memory so that they can be output grouped by the check type. When support for JSON output format was added the in-memory representation of each finding became more space consuming. As a result on huge ZIM-files with lots of errors zimcheck's memory consumption can be prohibitively large. For example, on a machine with 8GB of RAM the following checks run on [stackoverflow.com_en_all_2022-05.zim](http://download.kiwix.org/zim/stack_exchange/stackoverflow.com_en_all_2022-05.zim) (which has more than 64M entries) ended like follows:

`zimcheck --redirect_loop` completed in 8.5 minutes using 1097 MiB of RAM (there were no errors)
`zimcheck --url_external` completed in ~2 hours using 1219 MiB of RAM and producing a 79MiB output
`zimcheck --url_internal --url_external` reached 6GiB of RAM usage in ~8 hours and was still far from completion. Forcing (with the help of a debugger) premature output produced a 745MiB report. The run was then interrupted.

I don't see why we shouldn't change the output format (and the implementation) so that errors/warnings can be output as soon as they are discovered.

Pranavjeet-Naidu:
fixes #311 and unblocks #464 

- the problem with the previous implementation was that all the error messages were accumulated inside the ReportMsg vector with bloated up, sometimes exceeded 6GB RAM usage. This also blocked implementation of the --report-all flag. 
- the alternative is to use a thread-safe stream logger inside zimcheck.
- the relevant tests have been modified, the hardcoded strings have been updated.
- dead code is removed ( Errorlogger::report() )

Possible cons: 
- have to tradeoff grouping the errors by category.
- "logs" array shows up before the "status" boolean in the json object.


src/zimcheck/zimcheck.cpp
```diff
@@ -280,7 +282,12 @@ int zimcheck(const Options& args)
 
             if ( enabled_tests.isEnabled(TestType::REDIRECT))
                 test_redirect_loop(archive, error);
-        }
+
+                error.endLogStream();
+        } else {
+                error.startLogStream();
+                error.endLogStream();
+            }
```
veloman-yunkan: Fix indentation (may be a tab-vs-spaces issue)

test/zimcheck-test.cpp
```diff
@@ -531,12 +531,11 @@ TEST(zimcheck, metadata_poorzimfile)
...
-      "  Missing mandatory metadata: Title" "\n"
-      "  Missing mandatory metadata: Description" "\n"
-      "  Missing mandatory metadata: Illustration_48x48@1" "\n"
-      "  Language must contain at least 3 characters" "\n"
-      "  Language doesn't match regex: ^\\w{3}(,\\w{3})*$" "\n"
+      "[ERROR] Metadata errors:\n  Missing mandatory metadata: Title\n"
+      "[ERROR] Metadata errors:\n  Missing mandatory metadata: Description\n"
+      "[ERROR] Metadata errors:\n  Missing mandatory metadata: Illustration_48x48@1\n"
+      "[ERROR] Metadata errors:\n  Language must contain at least 3 characters\n"
+      "[ERROR] Metadata errors:\n  Language doesn't match regex: ^\\w{3}(,\\w{3})*$\n"
```
veloman-yunkan: The new format doesn't look right. Before "Metadata errors" (in plural) made sense because more than one errors could be output under that line. Now each line is followed by a single error (on a separate line).

This concern applies to other error types as well.

veloman-yunkan (commented):
Regarding the commit messages - refactoring is supposed to mean non-functional changes, i.e. refactoring should not have user-observable effects (or they should be considered negligible). In this case, the major change in the output structure doesn't fit that definition.

I think we should first define what the output in non-JSON case must look like for this new way of error reporting. Please provide a general outline of the order in which different errors are generated, so that @kelson42, @benoit74 and/or @rgaudin can propose a user friendly report format according to their preferences.

Pranavjeet-Naidu:
> Regarding the commit messages - refactoring is supposed to mean non-functional changes, i.e. refactoring should not have user-observable effects (or they should be considered negligible). In this case, the major change in the output structure doesn't fit that definition.

Yeah that's pretty fair. Given that it's more of an Optimization, I'll change the first commit message to this: 

```bash
optimize: modify logging from batching to streaming
```
(the other two commit's messages should remain the same right?)

>I think we should first define what the output in non-JSON case must look like for this new way of error reporting. Please provide a general outline of the order in which different errors are generated, so that [[@kelson42](https://github.com/kelson42)](https://github.com/kelson42), [[@benoit74](https://github.com/benoit74)](https://github.com/benoit74)and/or [[@rgaudin](https://github.com/rgaudin)](https://github.com/rgaudin) can propose a user friendly report format according to their preferences.

Yeah I agree, printing 'errors:' (plural) followed by a single error looks weird now that we aren't grouping them together. 
The easiest and cleanest way to handle streaming errors seems like flattening them into a single line. 
this is what a general outline of the errors ordering looks like: 
- Global Checks (Sequential): Integrity -> Checksum -> Metadata -> Favicon -> Main Page.
- Article Checks (Interleaved/Streaming): As `zimcheck` iterates through the content, errors for **Empty Articles**, **Internal URLs**, **External URLs**, and **Redundancies** are emitted dynamically the moment they are encountered.
- Post-Article Checks (Sequential):Redirect Loops.


**How it used to look (Batched):**
```text
[ERROR] Metadata errors:
  Missing mandatory metadata: Title
  Missing mandatory metadata: Description
```

**How it looks right now (basic streaming):**
```text
[ERROR] Metadata errors:
  Missing mandatory metadata: Title
[ERROR] Metadata errors:
  Missing mandatory metadata: Description
```

**Proposed New Format (Flat streaming):**
basically print them inline with the expansion
```text
[ERROR] Metadata: Missing mandatory metadata: Title
[ERROR] Metadata: Missing mandatory metadata: Description
[ERROR] Internal URL: A/non_existent.html not found in dangling_link.html
[WARNING] Empty Link: Found 1 empty link in article: empty_link.html
```

Does this single-line stream format look good to you for standard terminal output? If you approve, I can easily adjust the `std::cout` logic in `checks.cpp` to flatten the strings!

Also, these lines from src/zimcheck/zimcheck.cpp seems interesting, it directly supports this implementation :
```bash
/* Now we want to avoid to loop on the tests but on the article.
             *
             * If we loop of the tests we will have :
             *
             * for (test: tests) {
             * for(article: articles) {
             * data = article->getData();
             * ...
             * }
             * }
             *
             * And so we will get several the data of an article (and so decompression and so).
             * By looping on the articles first, we have :
             *
             * for (article: articles) {
             * data = article->getData();
             * for (test: tests) {
             * ...
             * }
             * }
             */

```

@veloman-yunkan @kelson42 @rgaudin @benoit74 tell me how this looks, sorry for the word vomit T_T

kelson42:
> Does this single-line stream format look good to you for standard terminal output? If you approve, I can easily adjust the `std::cout` logic in `checks.cpp` to flatten the strings!

@Pranavjeet-Naidu Yes, please go ahead.

src/zimcheck/zimcheck.cpp
Pranavjeet-Naidu: It was a tab vs spaces issue, should be correct now

Pranavjeet-Naidu:
@kelson42, I rebased the changes into the 3 phase commit thats going on right now, I can squash them if needed as well, thank you :)

src/zimcheck/checks.cpp
```diff
@@ -170,10 +175,16 @@ void ErrorLogger::setTestResult(TestType type, bool status) {
 
 void ErrorLogger::addMsg(MsgId msgid, const MsgParams& msgParams)
 {
+  std::lock_guard<std::mutex> lock(this->print_mutex);
```
veloman-yunkan: Such a mutex seems to have been needed since multithreaded processing has been introduced. If that's really a bug, then it has to be fixed in a separate commit.

src/zimcheck/checks.cpp
```diff
@@ -230,6 +221,20 @@ bool ErrorLogger::overallStatus() const {
...
+        jsonOutputStream << JSON::property("logs", JSON::startArray);
+        logStreamOpen = true;
+    }
+}
+
+void ErrorLogger::endLogStream() {
+    if ( jsonOutputStream.enabled() && logStreamOpen ) {
+        jsonOutputStream << JSON::endArray;
+        logStreamOpen = false;
+    }
```
veloman-yunkan: What will happen if `addInfo()` is called between `startLogStream()` and `endLogStream()`?

test/zimcheck-test.cpp
```diff
@@ -503,16 +503,14 @@ TEST(zimcheck, json_goodzimfile)
...
-      "[ERROR] Invalid checksum:" "\n"
-      "  ZIM Archive Checksum in archive: 00000000000000000000000000000000" "\n"
-      "" "\n"
-      "[INFO] Overall Test Status: Fail" "\n"
-      "[INFO] Total time taken by zimcheck: <3 seconds." "\n"
+      "[INFO] Verifying Internal Checksum...\n"
+      "  [ERROR] Wrong Checksum in ZIM archive\n"
+      "[ERROR] Checksum: ZIM Archive Checksum in archive: 00000000000000000000000000000000\n\n"
+      "[INFO] Overall Test Status: Fail\n"
+      "[INFO] Total time taken by zimcheck: <3 seconds.\n"
```
veloman-yunkan: The newline character on the end of every line was singled out intentionally. Please preserve that format/convention for expected output data (for all changes in this file).

test/zimcheck-test.cpp
```diff
@@ -670,16 +661,15 @@ TEST(zimcheck, external_url_check_poorzimfile)
...
-      "[INFO] Verifying Articles' content..." "\n"
-      "[INFO] Searching for redundant articles..." "\n"
-      "  Verifying Similar Articles for redundancies..." "\n"
-      "[WARNING] Redundant data found:" "\n"
-      "  article1.html and redundant_article.html" "\n"
-      "[INFO] Overall Test Status: Pass" "\n"
-      "[INFO] Total time taken by zimcheck: <3 seconds." "\n"
+      "[INFO] Verifying Articles' content...\n"
+      "[INFO] Searching for redundant articles...\n"
+      "  Verifying Similar Articles for redundancies...\n"
```
veloman-yunkan: This looks like a deviation from the adopted format.

test/zimcheck-test.cpp
```diff
@@ -719,51 +705,39 @@ TEST(zimcheck, redirect_loop_poorzimfile)
...
+      "[ERROR] Metadata: Language doesn't match regex: ^\\w{3}(,\\w{3})*$\n"
+      "[INFO] Searching for Favicon...\n"
+      "[ERROR] Favicon: Favicon is missing\n"
+      "[INFO] Searching for main page...\n"
+      "[ERROR] Main Page: Main Page Index stored in Archive Header: 4294967295\n"
+      "[INFO] Verifying Articles' content...\n"
+      "[ERROR] Internal URL: The following links:\n"
+      "- A/non_existent.html\n"
+      "(A/non_existent.html) were not found in article dangling_link.html\n"
+      "[ERROR] Empty Article: Entry empty.html is empty\n"
```
veloman-yunkan: This doesn't look nice. And it even looks wrong!

src/zimcheck/zimcheck.cpp
```diff
@@ -175,8 +174,6 @@ int zimcheck(const Options& args)
         } else if (arg.first == "--redirect_loop" && arg.second.asBool()) {
             enabled_tests.enable(TestType::REDIRECT);
             no_args = false;
-        } else if (arg.first == "--details") {
-            error_details = arg.second.asBool();
```
veloman-yunkan, line 179: So the `--details` option had lost its meaning a while ago. Getting rid of it deserves a separate commit (before the main changes of this PR are made) with an explanation. The `USAGE` string must be updated too.

veloman-yunkan (changes requested):
Fixing unit-tests is not a refactoring.

When a non-functional change breaks the unit-tests (and they are going to be fixed in a separate commit) that must be explicitly documented in the commit message.

src/zimcheck/checks.cpp
veloman-yunkan: The fix (if indeed needed) must come before the main changes of this PR.

src/zimcheck/checks.cpp
Pranavjeet-Naidu: It's going to break the json array that startLogStream will open, so probably output broken json.

src/zimcheck/checks.cpp
Pranavjeet-Naidu: I do feel like the mutex lock is needed, atleast that's what my first instinct is.

Also, Does this commit chain look fine: 
```
fix: add mutex lock to ErrorLogger
cli: remove deprecated --details option and update USAGE
optimize: modify logging from batching to streaming
test: update unit tests to match new streaming output
refactor: remove dead code
```

test/zimcheck-test.cpp
Pranavjeet-Naidu: Will be handled.

src/zimcheck/checks.cpp
veloman-yunkan: Fool-proofness (in the form of an `assert()` or something similar) must be added

Pranavjeet-Naidu: aren't assert statements only for Debug builds, iirc the compiler drops them for *Release*, maybe I'll add a hard exception instead.

src/zimcheck/checks.cpp
veloman-yunkan: - `and update USAGE` is redundant. Instead mention `zimcheck` since there are other tools besides in this repo.
- The last three commits should be squashed into one

test/zimcheck-test.cpp
Pranavjeet-Naidu: Yes, I missed it here, will handle it in the last commit.

src/zimcheck/checks.cpp
veloman-yunkan: `assert()`s are for catching bugs. Calling `addInfo()` while `startLogStream()` is in effect would mean a bug.

test/zimcheck-test.cpp
Pranavjeet-Naidu:
```cpp
{ MsgId::DANGLING_LINKS, { TestType::URL_INTERNAL, "Link '{{&normalized_link}}' was not found in article '{{&path}}'" } }
```
is this better?

veloman-yunkan: The non-normalized versions of the link were lost

Pranavjeet-Naidu: do you mean the array of [{{#links}} {{&value}} {{/links}}]? Oh yes, this should work then: 
```cpp
{ MsgId::DANGLING_LINKS, { TestType::URL_INTERNAL, "The following link(s): {{#links}}'{{&value}}' {{/links}}resolving to '{{&normalized_link}}' were not found in article '{{&path}}'" } }
```

veloman-yunkan: I don't like having a list of links on a single line. Maybe we should change the way those links are reported (and then it will have to be done as a separate preparatory change).

Pranavjeet-Naidu: Ah then how do I approach this? Make it a new pr or will a different commit suffice?

Pranavjeet-Naidu: @veloman-yunkan let me know if you'd like me to make it a different pr, I can continue with the current pr while  reverting these lines and the corresponding tests. The new pr will have the changes once we change how the links are reported in the first place.

veloman-yunkan: A separate commit in this PR will suffice.

A PR should 
1. start with as much groundwork/preparatory changes as possible,
2. then proceed to implementing the main objective in the simplest way imaginable,
3. optionally followed by additional clean-up enabled by the changes made on step 2 (though one should not artificially defer tiny code clean-up changes into standalone commits if they naturally fit into modifications carried out in step 2).

Each change/commit should be reasonably small and easy to formulate/describe and understand.

Pranavjeet-Naidu:
@veloman-yunkan, the changes are now up. Really sorry for the delay, I have my university finals going on right now. 
All the changes we discussed above have been addressed: 
- mutex lock as a separate commit
- removing the details flag in a separate commit
- the list of links format has been modified, the indentation should help now
- addInfo() now has assert guards, the tests should have the "/n" preserved as well.

kelson42:
@veloman-yunkan Can you please make a new review pass?

src/zimcheck/checks.h
```diff
@@ -83,6 +84,8 @@ class ErrorLogger {
     // testStatus[i] corresponds to the status of i'th test
     std::bitset<size_t(TestType::COUNT)> testStatus;
 
+    std::mutex print_mutex;
```
veloman-yunkan: 1. `snake_case` looks out of place among code using `camelCase` convention.
2. At this point, the mutex guards adding messages rather than printing them. `msgMutex` is a better name that may persist unmodified through the main change brought by this PR.

src/zimcheck/checks.cpp
```diff
@@ -50,8 +50,7 @@ std::unordered_map<MsgId, MsgInfo> msgTable = {
   { MsgId::MAIN_PAGE,        { TestType::MAIN_PAGE, "Main Page Index stored in Archive Header: {{&main_page_index}}" } },
   { MsgId::EMPTY_ENTRY,      { TestType::EMPTY, "Entry {{&path}} is empty" } },
   { MsgId::OUTOFBOUNDS_LINK, { TestType::URL_INTERNAL, "{{&link}} is out of bounds. Article: {{&path}}" } },
-  { MsgId::DANGLING_LINKS,   { TestType::URL_INTERNAL, "The following links:\n{{#links}}- {{&value}}\n{{/links}}({{&normalized_link}}) were not found in article {{&path}}" } },
-  { MsgId::EXTERNAL_LINK,    { TestType::URL_EXTERNAL, "{{&link}} is an external dependence in article {{&path}}" } },
+  { MsgId::DANGLING_LINKS,   { TestType::URL_INTERNAL, "Missing link(s) in article '{{&path}}':\n{{#links}}  - '{{&value}}' (resolves to '{{&normalized_link}}')\n{{/links}}" } },  { MsgId::EXTERNAL_LINK,    { TestType::URL_EXTERNAL, "{{&link}} is an external dependence in article {{&path}}" } },
```
veloman-yunkan: 1. The two separate lines for `MsgId::DANGLING_LINKS` and `MsgId::EXTERNAL_LINK` were accidentally merged into one line.
2. "Missing link(s)" doesn't sound right. The link is there, so it is definitely not missing. Please change to "Dangling link(s)".
3. I am a little concerned that the "**(resolves to** *normalized_link* **)**" part is repeated for all reported links, however I can't figure out a good way of outputting the normalized version of the link (shared for all links reported in this batch) only once.

src/zimcheck/checks.cpp
```diff
@@ -163,17 +169,36 @@ void ErrorLogger::infoMsg(const std::string& msg) const {
...
+}
+
 void ErrorLogger::setTestResult(TestType type, bool status) {
     testStatus[size_t(type)] = status;
 }
 
 void ErrorLogger::addMsg(MsgId msgid, const MsgParams& msgParams)
 {
-    std::lock_guard<std::mutex> lock(this->print_mutex);
+  std::lock_guard<std::mutex> lock(this->print_mutex);
```
veloman-yunkan: A clean PR should avoid introducing a defect in one commit and fixing it in a later commit.

veloman-yunkan (changes requested):

src/zimcheck/checks.cpp
Pranavjeet-Naidu: 1. Have separated them now
2. Thanks for this, have changed it. 
3. The only possible change I can think of is having that part at the header, so it will look like this: 
```
[ERROR] Internal URL: Dangling link(s) in article 'dangling_link.html' (resolve to 'A/non_existent.html'):
  - 'A/non_existent.html'
  - '../A/non_existent.html'
```

Pranavjeet-Naidu:
once you're happy with the current status, i can squash/rebase the changes @veloman-yunkan 
@kelson42 the ci fail seems to be unrelated to the fixes, I could be wrong but can you take a look please, thank you.

veloman-yunkan:
> once you're happy with the current status, i can squash/rebase the changes @veloman-yunkan

I am happy. Please clean-up the PR by pushing the changes of the last commit into the respective commits where they should belong. This job would be easier if you had addressed the review comments as separate fixup commits (via `git commit --fixup`) - then `git rebase -i --autosquash main` would handle them automatically.  

> @kelson42 the ci fail seems to be unrelated to the fixes, I could be wrong but can you take a look please, thank you.

The CI was affected by the upgrade of build platforms in the kiwix/kiwix-build repository (kiwix/kiwix-build#943, kiwix/kiwix-build#936). Now all openzim/kiwix C++ projects must have their CI builiders upgraded too.

Pranavjeet-Naidu:
> > once you're happy with the current status, i can squash/rebase the changes @veloman-yunkan
> 
> I am happy. Please clean-up the PR by pushing the changes of the last commit into the respective commits where they should belong. This job would be easier if you had addressed the review comments as separate fixup commits (via `git commit --fixup`) - then `git rebase -i --autosquash main` would handle them automatically.
> 
> > @kelson42 the ci fail seems to be unrelated to the fixes, I could be wrong but can you take a look please, thank you.
> 
> The CI was affected by the upgrade of build platforms in the kiwix/kiwix-build repository ([kiwix/kiwix-build#943](https://github.com/kiwix/kiwix-build/pull/943), [kiwix/kiwix-build#936](https://github.com/kiwix/kiwix-build/pull/936)). Now all openzim/kiwix C++ projects must have their CI builiders upgraded too.

The changes are up, don't mind the multiple force pushes, that was a fumble from my side.

veloman-yunkan (approved):
Please

PR #505, opened 2026-06-09
zimcheck: default full URL reporting + --quick

### LINKED ISSUE #464: zimcheck: report all links in error in HTML files

Pranavjeet-Naidu:
Fixes #464

**Breaking change:** `-U`, `-X`, `-A` (and the default no-args behavior) now report all dangling internal links and all external `src` errors in each HTML article. Previously, scanning stopped after the first error of each type per article.

- Added `-Q`/`--quick` to restore the previous faster behavior (first dangling internal link and/or first external `src` error per article). JSON output includes `"quick": true` when quick mode is used.

## Implementation

- Replaced the report-all flags with a single `quick` flag in `LinkReportOptions` (default: `false`).
- Early exit in link checks now happens only when `quick` is enabled.
- Added test fixtures and coverage for both full reporting and `-Q`/`--quick`.


Pranavjeet-Naidu:
@kelson42 @veloman-yunkan let me know if you'd like anything changed

kelson42 (commented):
@Pranavjeet-Naidu Thank you for your patience, I finally come back with a few comments. These comments should partly have been done in the issue already, sorry for that too.

My remarks:
* I don't remember the reason why it has been decided to only report one per page error of these two kinds. To me it sounds a bad decision. Default behaviour is buggy, and default behaviour should be to search and show all the errors (so no additional option needed of the kind you have introduced).
* I could understand that we need a "--quick" feature just to identify if there is a problem without making the full report.... so only one link per page... then please just introduce the new `--quick` which would imply just stopping further checks if one URL related problem is detected.

Pranavjeet-Naidu:
> @Pranavjeet-Naidu Thank you for your patience, I finally come back with a few comments. These comments should partly have been done in the issue already, sorry for that too.

Hi @kelson42, that's alright


> * I don't remember the reason why it has been decided to only report one per page error of these two kinds. To me it sounds a bad decision. Default behaviour is buggy, and default behaviour should be to search and show all the errors (so no additional option needed of the kind you have introduced).

Yeah that does seem like the right thing to do, I've modified the pr title as well, I'll work on it right now. 
> * I could understand that we need a "--quick" feature just to identify if there is a problem without making the full report.... so only one link per page... then please just introduce the new `--quick` which would imply just stopping further checks if one URL related problem is detected.

Yes, that seems right. Please take a look at the PR description and tell me if thats what we want now.

kelson42:
@Pranavjeet-Naidu The description should only include what is different with `main` branch, so I have simplified it, but what you have written seems correct

kelson42 (changes requested):
Please add option `-Q` for quick

For the rest, I think we are good to go with the code review @veloman-yunkan

kelson42 (approved):
LGTM from a user perspective

src/zimcheck/zimcheck.cpp
```diff
@@ -203,10 +207,17 @@ int zimcheck(const Options& args)
...
+        if (report_options.quick)
+            error.addInfo("quick", true);
         error.addInfo("file_name",  filename);
         error.infoMsg("[INFO] Checking zim file " + filename);
         error.infoMsg("[INFO] Zimcheck version is " + std::string(VERSION));
 
+        if (report_options.quick
+            && !enabled_tests.isEnabled(TestType::URL_INTERNAL)
+            && !enabled_tests.isEnabled(TestType::URL_EXTERNAL))
+            error.infoMsg("[WARNING] -Q/--quick has no effect without --url_internal or --url_external");
```
veloman-yunkan: Do we really need this? If yes, then why don't you merge the second `if (report_options.quick && ..)` into the first  `if (report_options.quick)` statement?

src/zimcheck/checks.h
```diff
@@ -50,6 +50,10 @@ class EnabledTests {
     bool isEnabled(TestType tt) const { return tests[size_t(tt)]; }
 };
 
+struct LinkReportOptions {
+  bool quick = false;
+};
```
veloman-yunkan: This is an unnecessarily narrow control. Why shouldn't we able to extend the `quick` mode to any check to which it may apply (rather than only to links)?

Please replace with

```c++
struct ZimCheckOptions {
  EnabledTests enabledTests;
  bool quick;
};
```

src/zimcheck/checks.cpp
```diff
@@ -296,11 +296,13 @@ class ArticleChecker
     typedef std::vector<html_link> LinkCollection;
 
 public: // functions
-    ArticleChecker(const zim::Archive& _archive, ErrorLogger& _reporter, ProgressBar& _progress, EnabledTests _checks)
+    ArticleChecker(const zim::Archive& _archive, ErrorLogger& _reporter, ProgressBar& _progress,
+                   EnabledTests _checks, LinkReportOptions _reportOptions)
```
veloman-yunkan: Introducing an extra parameter doesn't look good.

If the type `EnabledTests` was named a little differently you could have just enhanced that type with the additional mode controlling how checks should be performed. But the easy way is to go with a `ZimCheckOptions` struct that has an `EnabledTests enabled_tests` field in addition to other options (of which `quick` is going to be the first new one).

test/data/zimfiles/create_test_zimfiles
```diff
@@ -67,19 +67,40 @@ make__poor__zim()
     cd "$tmpdir"/small_zimfile_data
     touch empty.html
     mv favicon.png image.png
-    sed -i -e 's!favicon.png!image.png!' main.html
+    sed -e 's!favicon.png!image.png!' main.html > main.html.tmp && mv main.html.tmp main.html
```
veloman-yunkan: ???

How is the new command different from `sed -i`? Or does your version of `sed` not support `-i`?

test/data/zimfiles/create_test_zimfiles
```diff
@@ -67,19 +67,40 @@ make__poor__zim()
...
-    command_set3="sed -e '6 a <meta http-equiv=refresh content=\"0;URL=redirect_loop2.html\">' main.html"
-    eval $command_set > redirect_loop3.html
+    insert_refresh_meta()
+    {
+      awk -v url="$1" 'NR==6{print; print "<meta http-equiv=refresh content=\"0;URL=" url "\">"; next}1' main.html
+    }
+    insert_refresh_meta redirect_loop.html > redirect_loop.html
+    insert_refresh_meta redirect_loop3.html > redirect_loop2.html
+    # Historically redirect_loop3.html used redirect_loop.html as target (see command_set typo).
+    insert_refresh_meta redirect_loop.html > redirect_loop3.html
```
veloman-yunkan: This change must be undone and a different way of dealing with the problem must be found.

The test code sets up for two different redirect loops:

1. Self-referencing redirect_loop.html
2. Redirect loop between redirect_loop2.html and redirect_loop3.html referencing each other.

The problem is that after openzim/libzim#1065 `libzim` doesn't produce ZIM files containing redirect loops. So rerunning the `create_test_zimfiles` script cannot generate a ZIM file on which redirect loop detection by `zimcheck` can be tested.

BTW, this change should have resulted in the changed output for the tests involving redirect loop detection. Why isn't that the case?

test/zimcheck-test.cpp
```diff
@@ -147,6 +147,7 @@ R"(Zimcheck checks the quality of a ZIM file.
  -R --redundant       Redundant data check
  -U --url_internal    URL check - Internal URLs
  -X --url_external    URL check - External URLs
+ -Q --quick           Stop after the first dangling internal and/or external src error per HTML article
```
veloman-yunkan: Let's change the description of this option to "Report at most one error of each type per ZIM entry"

test/data/zimfiles/create_test_zimfiles
```diff
@@ -67,19 +67,40 @@ make__poor__zim()
...
+    cat > multi_external_src.html <<'EOF'
+<!DOCTYPE html>
+<html>
+  <head><meta charset="UTF-8"><title>multi external</title></head>
+  <body>
+    <img src="http://a.io/one.png">
+    <img src="http://a.io/two.png">
+  </body>
+</html>
+EOF
```
veloman-yunkan: It looks like it could be better to add a couple of HTML files in `small_zimfile_data` and update the `make__good__zim()` so that they aren't included in `good.zim`. Such files can be located under a new subdirectory `faulty_articles/`.

veloman-yunkan (changes requested):

src/zimcheck/zimcheck.cpp
Pranavjeet-Naidu: Felt like it might be relevant in case someone makes a typo of having -Q without any URL flags, I'll remove it for now.

test/data/zimfiles/create_test_zimfiles
Pranavjeet-Naidu: sed -i on BSD/macOS can leave a main.html-e backup that zimwriterfs then picks up as an extra entry. The redirect pattern avoids that and works the same on Linux CI and macOS dev machines.

seemed relevant : https://stackoverflow.com/questions/16745988/sed-command-with-i-option-in-place-editing-works-fine-on-ubuntu-but-not-mac/16746032

test/data/zimfiles/create_test_zimfiles
Pranavjeet-Naidu: > This change must be undone and a different way of dealing with the problem must be found.

I've reverted the awk workaround.

> The test code sets up for two different redirect loops:
Self-referencing redirect_loop.html
Redirect loop between redirect_loop2.html and redirect_loop3.html referencing each other.
The problem is that after https://github.com/openzim/libzim/pull/1065 libzim doesn't produce ZIM files containing redirect loops. So rerunning the create_test_zimfiles script cannot generate a ZIM file on which redirect loop detection by zimcheck can be tested.

would this work?
A one-time build with pre-#1065 libzim + fixed script to obtain the correct redirect graph, then stop relying on script generation going forward. There's two ways I can think of to make this work: 

1. Regenerate `poor.zim` with the correct graph and document that redirect-loop entries are frozen
2. Add `redirect_loop.zim` as a dedicated fixture so `poor.zim` stays fully regeneratable with current libzim 
Which approach do you prefer think would be better. 

> BTW, this change should have resulted in the changed output for the tests involving redirect loop detection. Why isn't that the case?

The CI uses the committed `poor.zim`, not live script output. The awk version preserved the same redirect graph as the original bug (`eval $command_set` used for `redirect_loop3.html` instead of `eval $command_set3`), so even a regen from HTML wouldn't necessarily change zimcheck output. If `poor.zim` were regenerated with libzim post-#1065, redirect loops would be stripped entirely and redirect-loop tests would fail.

kelson42:
@veloman-yunkan Can we please move forward this review?

veloman-yunkan:
@Pranavjeet-Naidu Please publish the latest version containing fixes to the last review comments. I will try to complete the part related to the test data.

Pranavjeet-Naidu:
> @Pranavjeet-Naidu Please publish the latest version containing fixes to the last review comments. I will try to complete the part related to the test data.

the merge conflicts should be fixed once your work on the test data is done then, thank you.

veloman-yunkan:
@Pranavjeet-Naidu Thanks for your work! I have completed this PR. While cleaning it up I rewrote all the commits, which resulted in your authorship being lost. You can restore it via a special rebase operation (by introducing before each commit a small change and its reversal and then squashing each such triplet into one commit). For example, a commit history (A, B, C) first has to be converted into

- do_something
- undo_it
- A
- do_something
- undo_it
- B
- do_something
- undo_it
- C

Then each of

- (do_something, undo_it, A)
- (do_something, undo_it, B)
- (do_something, undo_it, C)

must be squashed into one commit with the commit message of A, B & C respectively.

Pranavjeet-Naidu:
> @Pranavjeet-Naidu Thanks for your work! I have completed this PR. While cleaning it up I rewrote all the commits, which resulted in your authorship being lost. You can restore it via a special rebase operation (by introducing before each commit a small change and its reversal and then squashing each such triplet into one commit). For example, a commit history (A, B, C) first has to be converted into
> 
> * do_something
> * undo_it
> * A
> * do_something
> * undo_it
> * B
> * do_something
> * undo_it
> * C
> 
> Then each of
> 
> * (do_something, undo_it, A)
> * (do_something, undo_it, B)
> * (do_something, undo_it, C)
> 
> must be squashed into one commit with the commit message of A, B & C respectively.

Thanks for this, learnt something new!

Issue #509, opened 2026-06-26
Zimcheck choke on article with double slash `//` in path


benoit74:
ZIM: [tests_en_doubleslash_2026-06.zim.zip](https://github.com/user-attachments/files/29374177/tests_en_doubleslash_2026-06.zim.zip)

The ZIM works well (at least in kiwix-apple and kiwix-serve) but zimcheck complains it found an error:

```bash
> python zim_doubleslash.py && zimcheck tests_en_doubleslash_2026-06.zim
INFO:tests_en_doubleslash:Building ZIM
Resolve redirect
set index
[INFO] Checking zim file tests_en_doubleslash_2026-06.zim
[INFO] Zimcheck version is 3.7.0
[INFO] Verifying ZIM-archive structure integrity...
[INFO] Avoiding redundant checksum test (already performed by the integrity check).
[INFO] Checking metadata...
[INFO] Searching for Favicon...
[INFO] Searching for main page...
[INFO] Verifying Articles' content...
[INFO] Searching for redundant articles...
  Verifying Similar Articles for redundancies...
[INFO] Checking for redirect loops...
[ERROR] Invalid internal links found:
  ../../acme is out of bounds. Article: foo//bar
[INFO] Overall Test Status: Fail
[INFO] Total time taken by zimcheck: <3 seconds.
```

```bash
> zimdump list tests_en_doubleslash_2026-06.zim 
acme
foo//bar
```

Python codebase to create the ZIM: https://gist.github.com/benoit74/e90a3111526ab55f4c61cbe4461a3c46

This is again a repro case for a real problem encountered in the wild at https://cms.openzim.org/title/73cbae7f-8ccb-4686-a512-a73ce78af0fb (and probably many others, we've just enabled blocking ZIM with zimcheck issues in the CMS)

kelson42:
This is a priority bug for the CMS

veloman-yunkan:
This doesn't have anything to do with the double slash. It works the same way if the path of the problematic article is changed from `foo//bar` to `foo/bar`.

Or did you mean that `foo//bar` should be interpreted as a file "bar" located in a subdirectory "" (empty string) of directory "foo" (thus allowing to ascend two levels via `../..`)?

```
- "foo"/
   - ""/
      - bar
```

benoit74:
> This doesn't have anything to do with the double slash. It works the same way if the path of the problematic article is changed from foo//bar to foo/bar.

[...]

PR #511, opened 2026-07-07
Overhaul of link normalization

### LINKED ISSUE #506: Zimcheck choke on link with `./` (dot followed by slash) when article is at a path with slash as well
ZIM: [tests_en_dotslash_2026-06.zim.zip](https://github.com/user-attachments/files/29349016/tests_en_dotslash_2026-06.zim.zip)

The ZIM works well (at least in kiwix-apple and kiwix-serve) but zimcheck complains it found an error:

```bash
> python zim_dotslash.py && zimcheck tests_en_dotslash_2026-06.zim
INFO:tests_en_dotslash:Building ZIM
Resolve redirect
set index
[INFO] Checking zim file tests_en_dotslash_2026-06.zim
[INFO] Zimcheck version is 3.7.0
[INFO] Verifying ZIM-archive structure integrity...
[INFO] Avoiding redundant checksum test (already performed by the integrity check).
[INFO] Checking metadata...
[INFO] Searching for main page...
[INFO] Verifying Articles' content...
[INFO] Searching for redundant articles...
  Verifying Similar Articles for redundancies...
[INFO] Checking for redirect loops...
  The following links:
- ../folder2./Nairobi
(folder2Nairobi) were not found in article folder1/home
[INFO] Overall Test Status: Fail
[INFO] Total time taken by zimcheck: <3 seconds.
```

```bash
> zimdump list tests_en_dotslash_2026-06.zim                      
folder1/home
folder2./Nairobi
```

Python codebase to create the ZIM:

https://gist.github.com/benoit74/dad4ecf8db9c8ae7ccf3a8b6913eac0f

This is a repro case for a real problem encountered in the wild at https://cms.openzim.org/title/73cbae7f-8ccb-4686-a512-a73ce78af0fb (and probably many others, we've just enabled blocking ZIM with zimcheck issues in the CMS)

### LINKED ISSUE #516: zimcheck must report internal links with absolute-path URLs
See https://github.com/openzim/zim-tools/pull/511#issuecomment-4950138577

veloman-yunkan:
Fixes #506
Fixes #516 and the bugs reported in #488 (overriding the latter PR), as well as a few other defects not reported elsewhere.

This is a complete overhaul of the `normalize_link()` utility:

- First of all it is renamed to `resolveLinkTarget()`.
- The semantics of its second parameter is changed so that its functionality is easier to describe based on [RFC3896][1] (now the second parameter is the ZIM path of the entry containing the link to be resolved, whereas before it had to be the "parent directory" of that entry).
- The bug with the detection and elimination of parent-directory (`../`) and same directory (`./`) entries in the URL was fixed (#506).
- Absolute path URLs are rejected (#516)
- Fragment and/or search components of the URL no longer influence out-of-bounds URL detection (a bug reported in #488)
- Handling the parent-directory (`../`) near the top level has changed to match how ZIM paths interact with URL-building. In general a ZIM path looks like a relative path (e.g. `a/b`) and can look like an absolute path only if a leading `/` is explicitly provided. Therefore, `/` is no longer considered a root path - it is possible to ascend one level higher to the empty path (i.e. "/.." resolves to "").
- Path resolution is performed on URI-decoded URL (for example in the old implementation `/%2e%2e/` and/or `/%2e/` present in the URL would escape dot-segment removal and would sneak into the output as `/../` and/or `/./`, respectively).

The behaviour of `resolveLinkTarget()` is defined based on [RFC3986][1] as follow:

> A ZIM-path `ZIMPATH` in a ZIM-file `ZIMFILE.zim` is considered as part of an imaginary full URL
> `zim://ZIMFILE.zim/ZIMPATH` (the `zim:` scheme can be alternatively replaced with `http:`, as if
> the content of the ZIM-file is served directly at `http://ZIMFILE.zim/`, `ZIMFILE.zim` playing the role
> of the host component). If a relative URL `RELURL` in the context of the resource at that ZIM-path
> resolves (following the rules of [RFC3986][1] with the exceptions listed below) to
> `zim://ZIMFILE.zim/RESOLVEDPATH` (or, for the alternative approach, to `http://ZIMFILE.zim/RESOLVEDPATH`)
> then the result of `resolveLinkTarget(RELURL, ZIMPATH)` is `RESOLVEDPATH`.
> 
> Differences from [RFC3986][1] are the following:
> - Attempts to ascend above the top level (when resolving parent directory (`../`) dot-segments) are treated as an error rather than as a no-op (an `OutOfBoundsURL` exception is thrown)
> - Absolute path URLs result in an `AbsolutePathURL` exception

Also
- `isOutofBounds()` has been dropped; instead out-of-bound URLs are reported by `resolveLinkTarget()` (via an exception).

[1]: https://www.rfc-editor.org/info/rfc3986/#section-5


veloman-yunkan:
I figured out that URI-decoding must be applied differently (only to the first argument). Fixing it now.

kelson42:
> I figured out that URI-decoding must be applied differently (only to the first argument). Fixing it now.

Like any stuff related to URL, critical and subtil. Solid unit testing is the only substainable way forward.

veloman-yunkan:
After fixing the noticed problem I discovered another potential one, and also started doubting if performing path clean-up after URI-decoding is even appropriate.

The newly discovered problem is that ZIM paths can be arbitrary. What if there is a ZIM entry with path `abc/../def/foo.html`? From ZIM's perspective a link to `bar.html` in that article should resolve to `abc/../def/bar.html`. But my implementation will eliminate the `/../` from it and return `def/bar.html`. The last commit demonstrates other pitfalls.

Back to URI decoding, what if we have a link `%2e%2e/bar.html` (which URL-decodes to `../bar.html` in ZIM article `/dir/subdir/foo.html`? Should it resolve to `/dir/subdir/%2e%2e/bar.html`, `/dir/subdir/../bar.html` or `/dir/bar.html`?

kelson42:
@veloman-yunkan The only thing I can help with: encoded reserved characters, like `..` or `/` are URL content and should not be "interpreted".

veloman-yunkan:
> @veloman-yunkan The only thing I can help with: encoded reserved characters, like `..` or `/` are URL content and should not be "interpreted".

@kelson42 The problem is that in URLs they are interpreted by the browsers automatically.

veloman-yunkan:
> Back to URI decoding, what if we have a link `%2e%2e/bar.html` (which URL-decodes to `../bar.html` in ZIM article `/dir/subdir/foo.html`? Should it resolve to `/dir/subdir/%2e%2e/bar.html`, `/dir/subdir/../bar.html` or `/dir/bar.html`?

Testing shows that the correct result (at least from the point of view of Firefox and Chromium) is `/dir/bar.html` (i.e. path resolution is performed on URI-decoded URL).

Also testing revealed that adjacent slashes in URLs must not be collapsed into one (i.e. `/dir//subdir` must stay as is), and empty "directory" names between adjacent slashes are implied during path resolution (as suggested as a possibility in my question in https://github.com/openzim/zim-tools/issues/509#issuecomment-4877023894).

veloman-yunkan:
Also found a subtlety with the inclusion of a trailing slash in the output of the URL after [removal of dot segments](https://www.rfc-editor.org/info/rfc3986/#section-5.2.4):

`/a/b/..` → `/a/` (trailing slash is present!)
`/a/b/../` → `/a/` (no extra trailing slash is added!)
`/a/b/..//` → `/a//` (an extra trailing slash is added!)
`/a/b/.` → `/a/b/` 
`/a/b/./` → `/a/b/`
`/a/b/.//` → `/a/b//`

Overall, need to make sure that all of [Reference Resolution Examples of RFC3986](https://www.rfc-editor.org/info/rfc3986/#section-5.4) are reflected in our tests. The one for `g:h` was completely overlooked!

veloman-yunkan:
> Also found a subtlety with the inclusion of a trailing slash in the output of the URL after [removal of dot segments](https://www.rfc-editor.org/info/rfc3986/#section-5.2.4):
> 
> `/a/b/..` → `/a/` (trailing slash is present!)
> `/a/b/../` → `/a/` (no extra trailing slash is added!)
> `/a/b/..//` → `/a//` (an extra trailing slash is added!)

An additional complication comes from the fact that paths of ZIM entries don't necessarily start with a `/`. Both `foo/bar` and `/foo/bar` are valid and different ZIM paths. In practice, ZIM paths are appended to the API endpoint URL, so with `kiwix-serve` those entries will be accessible, for example, as `/content/ZIMNAME/foo/bar` and `/content/ZIMNAME//foo/bar`.

So it must be defined whether our URL normalization utility works in the scope of ZIM paths or URLs constructed out of those paths.

veloman-yunkan:
A problem arises when an absolute link is involved. Before writing the previous comment, I assumed that in the context of the unit tests `/foo/bar` was meant to refer to the entry `foo/bar` (or, maybe, `/foo/bar`) in the current ZIM file. And it is only now that I figured out that from the browser's point of view the URL `/foo/bar` found in the HTML fetched from, say, `http://localhost:8080/content/test_zim_file/welcome.html` resolves to `http://localhost:8080/foo/bar` and should therefore be considered an out-of-bounds URL!

Therefore a natural question is - are absolute URLs in the form of paths (i.e. without the scheme & host components) allowed in ZIMs (assuming no URL rewriting in the front-end)?

@kelson42 @rgaudin @benoit74

kelson42:
> A problem arises when an absolute link is involved.

URL must all be relative, this is in the specification https://wiki.openzim.org/wiki/Article_Format

Absolute URL should be reported similar to broken URL

kelson42:
> > @veloman-yunkan The only thing I can help with: encoded reserved characters, like `..` or `/` are URL content and should not be "interpreted".
> 
> @kelson42 The problem is that in URLs they are interpreted by the browsers automatically.

I don't think this is the point.

Check https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05/%2F vs https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05//.

veloman-yunkan:
> > A problem arises when an absolute link is involved.
> 
> URL must all be relative, this is in the specification https://wiki.openzim.org/wiki/Article_Format
> 
> Absolute URL should be reported similar to broken URL



> > > @veloman-yunkan The only thing I can help with: encoded reserved characters, like `..` or `/` are URL content and should not be "interpreted".
> > 
> > 
> > @kelson42 The problem is that in URLs they are interpreted by the browsers automatically.
> 
> I don't think this is the point.
> 
> Check https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05/%2F vs https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05//.

@kelson42 Your example has a weakness - in the above URLs the URI encoding is in the fragment component and handling of those URLs involves the viewer JS code to which any observed difference could be attributed. But I made sure that direct content access versions of the above URLs indeed lead to different pages when typed in the browser address bar:

- https://browse.library.kiwix.org/content/wikipedia_fr_all_maxi_2026-05/%2F
- https://browse.library.kiwix.org/content/wikipedia_fr_all_maxi_2026-05//

However with URI-encoded dots that's not the case. Try

- https://browse.library.kiwix.org/content/wikipedia_en_all_maxi_2026-02/nonexistent/../Dot
- https://browse.library.kiwix.org/content/wikipedia_en_all_maxi_2026-02/nonexistent/%2E%2E/Dot

both of which resolve to https://browse.library.kiwix.org/content/wikipedia_en_all_maxi_2026-02/Dot (and that is done by the browser, which can be verified by copying the link). BTW, typing those URLs in the address bar directly still doesn't suppress having them resolved before sending the request to the server.

veloman-yunkan:
> > A problem arises when an absolute link is involved.
> 
> URL must all be relative, this is in the specification https://wiki.openzim.org/wiki/Article_Format
> 
> Absolute URL should be reported similar to broken URL

@kelson42 Thanks for the info. The old code suggested that absolute URLs were valid. I will now start reporting them as errors.

kelson42:
> > > A problem arises when an absolute link is involved.
> 
> > 
> 
> > URL must all be relative, this is in the specification https://wiki.openzim.org/wiki/Article_Format
> 
> > 
> 
> > Absolute URL should be reported similar to broken URL
> 
> 
> 
> @kelson42 Thanks for the info. The old code suggested that absolute URLs were valid. I will now start reporting them as errors.

It was the case indeed, but this is something like 15 years ago and before...

veloman-yunkan:
The difference in the handling of the percent-encoded slash and dot characters [noted above](https://github.com/openzim/zim-tools/pull/511#issuecomment-4966992558) is that the former is a reserved character for general URI syntax, while the latter is not. According to [RFC 3986](https://www.rfc-editor.org/info/rfc3986/#section-2.3):

> URIs that differ in the replacement of an unreserved character with its corresponding percent-encoded US-ASCII octet are equivalent: they identify the same resource.  However, URI comparison implementations do not always perform normalization prior to comparison (see [Section 6](https://www.rfc-editor.org/info/rfc3986/#section-6)).  For consistency, percent-encoded octets in the ranges of ALPHA (%41-%5A and %61-%7A), DIGIT (%30-%39), hyphen (%2D), period (%2E), underscore (%5F), or tilde (%7E) should not be created by URI producers and, when found in a URI, should be decoded to their corresponding unreserved characters by URI normalizers.

benoit74:
Since the way (most) web browsers works prevent accessing a ZIM item either with `/./` or `/../` in its path or starting with `./` or `../`, I would like zimcheck to report an error if such an item is found. 

This does not have to be part of the specification, but we now know this cannot work well with kiwix-serve (and probably other readers as well) given the RFC you've identified.

And this opens the door to not over-complication for this PR.

Regarding ZIM entry which path starts with a `/`, I feel like they are indeed valid, so IMHO we should allow such situations. E.g. a link to `.//bar` from an item at `foo` path should resolve to `/bar` path and be considered as OK if there is a ZIM item there.

Note that kiwix-android and kiwix-pwa and kiiwx-serve without viewer already achieves to serve such items. kiwix-apple and kiwix-serve with viewer have a bug, I think we should fix these issues.

[tests_en_slash_2026-07.zim.zip](https://github.com/user-attachments/files/30034856/tests_en_slash_2026-07.zim.zip)

veloman-yunkan:
> Since the way (most) web browsers works prevent accessing a ZIM item either with `/./` or `/../` in its path or starting with `./` or `../`, I would like zimcheck to report an error if such an item is found.
> 
> This does not have to be part of the specification, but we now know this cannot work well with kiwix-serve (and probably other readers as well) given the RFC you've identified.

@kelson42 Does your reaction to @benoit74's suggestion mean that you formally approve it? If yes, should we have a `zimcheck` option for that new check?

veloman-yunkan:
> > > @veloman-yunkan The only thing I can help with: encoded reserved characters, like `..` or `/` are URL content and should not be "interpreted".
> > 
> > 
> > @kelson42 The problem is that in URLs they are interpreted by the browsers automatically.
> 
> I don't think this is the point.
> 
> Check https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05/%2F vs https://browse.library.kiwix.org/viewer#wikipedia_fr_all_maxi_2026-05//.

BTW, it looks like the difference between how those two URLs are handled rather comes from the reverse proxy (see kiwix/operations#725). Providing the extra slash in URI-encoded form circumvents the redirection by reverse proxy so that the URL is reached to the `kiwix-serve`. I believe that without any intervention by the reverse proxy those URLs would be handled in the same way by `kiwix-serve`.

kelson42:
> > Since the way (most) web browsers works prevent accessing a ZIM item either with `/./` or `/../` in its path or starting with `./` or `../`, I would like zimcheck to report an error if such an item is found.
> > 
> > This does not have to be part of the specification, but we now know this cannot work well with kiwix-serve (and probably other readers as well) given the RFC you've identified.
> 
> @kelson42 Does your reaction to @benoit74's suggestion mean that you formally approve it? If yes, should we have a `zimcheck` option for that new check?

Before anything else, I request a dedicated issue about this impossibility to request via HTTP a few very specific path at which a page might be stored in the ZIM.

This issue should:
* list the concerned strings
* Explain why this is impossible
* Explicit current impact for each of the supported readers

benoit74:
I've opened https://github.com/openzim/overview/issues/90 to track the effort on inaccessible paths.

veloman-yunkan:
> Overall, need to make sure that all of [Reference Resolution Examples of RFC3986](https://www.rfc-editor.org/info/rfc3986/#section-5.4) are reflected in our tests. The one for `g:h` was completely overlooked!

`g:h` must not make it to `resolveLinkTarget()` since it must not be categorized as an internal link (will be addressed in a different PR).

veloman-yunkan:
@kelson42 I have finished working on this PR.

kelson42 (approved):
I have made a superficial review of the code and have seen nothing which seems to be a problem at my eyes.

@veloman-yunkan Thank you very much for taking this - not that easy problem - very rigorously.

PR #515, opened 2026-07-11
Improve the refresh URL parsing

### LINKED ISSUE #513: zimwriterfs interprets `'` in meta-refresh URL field literally
Example: `./Birth_of_Elizabeth_II` (modified to remove section ids)

```html
<html>
  <head>
    <title>Birth of Elizabeth II</title>
    <meta http-equiv="refresh" content="0;URL='./Elizabeth_II'" />
  </head>
  <body>
    <a href="./Elizabeth_II">Birth of Elizabeth II</a>
  </body>
</html>
```

zimwriterfs output:

```
zimwriterfs: 'Birth_of_Elizabeth_II' HTML redirection target path 'zim/'./Elizabeth_II'' doesn't exist.
```

Removing the `'` around `./Elizabeth_II` causes the error to go away.

kelson42:
Best parsing of the HTML HEAD refresh URL (using regex) + proper testing

Fixes #513

### DIFF
--- a/src/zimwriterfs/tools.cpp
+++ b/src/zimwriterfs/tools.cpp
@@ -26,11 +26,11 @@
 #include <fstream>
 #include <iostream>
 #include <iomanip>
+#include <regex>
 #include <map>
 
 #include <zlib.h>
 #include <magic.h>
-
 
 static const std::map<std::string, std::string> extMimeTypes = {
     {"html",       "text/html"},
@@ -153,6 +153,21 @@ std::string getFileContent(const std::string& path)
   throw(errno);
 }
 
+std::string extractRedirectUrlFromHtmlHeadValue(const std::string& refresh_value)
+{
+  std::string url;
+  static const std::regex refresh_regex(R"(^\s*([0-9.]+)\s*(?:[;,]\s*(?:url *=)?\s*(["']?)(.*?)\2)?\s*$)", std::regex::icase);
+
+  std::smatch match;
+  if (std::regex_match(refresh_value, match, refresh_regex)) {
+    if (match[3].matched && !match[3].str().empty()) {
+      url = match[3].str();
+    }
+  }
+
+  return url;
+}
+
 std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
 {
   std::string url;
@@ -169,16 +184,10 @@ std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
           if ((attribute
                = gumbo_get_attribute(&child->v.element.attributes, "content"))
               != NULL) {
-            std::string targetUrl = attribute->value;
-            std::size_t found = targetUrl.find("URL=") != std::string::npos
-                                    ? targetUrl.find("URL=")
-                                    : targetUrl.find("url=");
-            if (found != std::string::npos) {
-              url = targetUrl.substr(found + 4);
-            } else {
+            url = extractRedirectUrlFromHtmlHeadValue(attribute->value);
+            if (url.empty()) {
               throw std::string(
-                  "Unable to find the redirect/refresh target url from the "
-                  "HTML DOM");
+                "Unable to find the redirect/refresh target URL in the HTML DOM");
             }
           }
         }
--- a/src/zimwriterfs/tools.cpp
+++ b/src/zimwriterfs/tools.cpp
@@ -186,7 +186,7 @@ std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
               != NULL) {
             url = extractRedirectUrlFromHtmlHeadValue(attribute->value);
             if (url.empty()) {
-              throw std::string(
+              throw std::runtime_error(
                 "Unable to find the redirect/refresh target URL in the HTML DOM");
             }
           }
--- a/src/zimwriterfs/tools.cpp
+++ b/src/zimwriterfs/tools.cpp
@@ -26,11 +26,11 @@
 #include <fstream>
 #include <iostream>
 #include <iomanip>
+#include <regex>
 #include <map>
 
 #include <zlib.h>
 #include <magic.h>
-
 
 static const std::map<std::string, std::string> extMimeTypes = {
     {"html",       "text/html"},
@@ -153,6 +153,21 @@ std::string getFileContent(const std::string& path)
   throw(errno);
 }
 
+std::string extractRedirectUrlFromHtmlHeadValue(const std::string& refresh_value)
+{
+  std::string url;
+  static const std::regex refresh_regex(R"(^\s*([0-9]+)\s*(?:;\s*url=\s*(["']?)(.*?)\2)?\s*$)", std::regex::icase);
+
+  std::smatch match;
+  if (std::regex_match(refresh_value, match, refresh_regex)) {
+    if (match[3].matched && !match[3].str().empty()) {
+      url = match[3].str();
+    }
+  }
+
+  return url;
+}
+
 std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
 {
   std::string url;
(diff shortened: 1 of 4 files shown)


src/zimwriterfs/tools.cpp
```diff
@@ -169,23 +184,19 @@ std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
...
           }
         }
       }
     }
   }
 
+  if (url.empty()) {
+    throw std::string(
+        "Unable to find the redirect/refresh target URL in the "
+        "HTML DOM");
```
veloman-yunkan: C++ allows to throw a value of any type, however it is customary to use a type derived from `std::exception`. `std::runtime_error` would be more appropriate here. I see that this is carried over from old code, however we shouldn't miss the opportunity to fix existing defects.

src/zimwriterfs/tools.cpp
```diff
@@ -153,6 +153,21 @@ std::string getFileContent(const std::string& path)
   throw(errno);
 }
 
+std::string extractRedirectUrlFromHtmlHeadValue(const std::string& refresh_value)
+{
+  std::string url;
+  static const std::regex refresh_regex(R"(^\s*([0-9]+)\s*(?:;\s*url=\s*(["']?)(.*?)\2)?\s*$)", std::regex::icase);
```
veloman-yunkan: Just for completeness, according to https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv:

- the `url=` part is optional (i.e. `0;http://example.com` is also a valid value for the `content` attribute of `<meta http-equiv`> tag and is equivalent to `0;url=http://example.com`)
- whitespace can be present between `url` and `=`, too
- `,` (in addition to `;`) is also a valid delimiter between the time value and the URL

src/zimwriterfs/tools.cpp
```diff
@@ -169,23 +184,19 @@ std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
...
         }
       }
     }
   }
 
+  if (url.empty()) {
+    throw std::string(
+        "Unable to find the redirect/refresh target URL in the "
+        "HTML DOM");
+  }
```
veloman-yunkan: Moving the exception raising statement out of the loop is wrong - it results in an error for every HTML not containing a `<meta http-equiv>` tag with a valid refresh URL (i.e. all HTML articles will fail to be processed by `zimwriterfs`). This bug introduced by this PR points to a missing test that checks `zimwriterfs` as a whole.

veloman-yunkan (changes requested):

src/zimwriterfs/tools.cpp
veloman-yunkan: I see that Copilot [had reported the same bug](https://github.com/openzim/zim-tools/pull/515#discussion_r3565079425)

src/zimwriterfs/tools.cpp
kelson42: Improvements implemented

src/zimwriterfs/tools.cpp
kelson42: Fixed

src/zimwriterfs/tools.cpp
kelson42: fixed

test/zimwriterfs-zimcreatorfs.cpp
```diff
@@ -178,3 +179,14 @@ TEST(ZimCreatorFSTest, ParseRedirect)
...
+
+
+TEST(ZimwriterfsTools, extractRedirectUrlFromHtmlHeadValue)
+{
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue("0;URL=foobar"), "foobar");
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue(" 0 ; URL=foobar "), "foobar");
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue(" 22 ;url=foobar "), "foobar");
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue(" 22 ,url ='foobar' "), "foobar");
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue("0;url=\"foobar\" "), "foobar");
+  EXPECT_EQ(extractRedirectUrlFromHtmlHeadValue("0;foobar"), "foobar");
```
veloman-yunkan: It is good practice when writing unit-tests to test for different return values. Your tests can be satisfied by a stupid/placeholder implementation like below (which is acceptable as a first step in TDD when just introducing a new function):

```c++
std::string extractRedirectUrlFromHtmlHeadValue(const std::string& refresh_value)
{
  return "foobar";
}
```

veloman-yunkan (approved):
Please squash all commits into one before merging.

veloman-yunkan:
Actually Copilot found a serious problem:

> **src/zimwriterfs/tools.cpp:191**
> 
>  * `extractRedirectUrlFromHtml()` now throws `std::runtime_error` when parsing fails, but the only caller (`ZimCreatorFS::parseAndAdaptHtml`) currently catches `std::string`. This changes behavior from “log and continue” to an uncaught exception path for malformed/unsupported refresh values. To preserve current behavior without requiring cross-file changes, throw a `std::string` here (as before) or update the caller in the same PR to catch `std::exception`.
> 
> 
> ```
>             url = extractRedirectUrlFromHtmlHeadValue(attribute->value);
>             if (url.empty()) {
>               throw std::runtime_error(
>                 "Unable to find the redirect/refresh target URL in the HTML DOM");
>             }
> ```

src/zimwriterfs/tools.cpp
```diff
@@ -169,16 +184,10 @@ std::string extractRedirectUrlFromHtml(const GumboVector* head_children)
...
-                                    : targetUrl.find("url=");
-            if (found != std::string::npos) {
-              url = targetUrl.substr(found + 4);
-            } else {
-              throw std::string(
-                  "Unable to find the redirect/refresh target url from the "
-                  "HTML DOM");
+            url = extractRedirectUrlFromHtmlHeadValue(attribute->value);
+            if (url.empty()) {
+              throw std::runtime_error(
```
veloman-yunkan, line 189: Please change the type of the thrown exception in a separate commit and fix the call site too

veloman-yunkan (changes requested):
Please address the bug caught by Copilot

src/zimwriterfs/tools.cpp
kelson42: fixed

test/zimwriterfs-zimcreatorfs.cpp
veloman-yunkan: @kelson42 I guess you missed the comment above

kelson42: I don't understand what you want here

kelson42:
@veloman-yunkan I have rebased and squashed and fixed the last pb reported by Copilot. For your remark regarding the testing, feel free to just add what you suggested, seems a small thing, and might just be quicker to do it yourself.

test/zimwriterfs-zimcreatorfs.cpp
veloman-yunkan: I want that something other than "foobar" appears as a second argument of `EXPECT_EQ()` in your tests. Otherwise without looking at the implementation one can suspect that the function `extractRedirectUrlFromHtmlHeadValue()` always returns "foobar". As a result the real implementation can also be affected by a bug making it work for the single value being tested, but failing for other cases. In short - diversify the testing of the return value too.

kelson42:
@veloman-yunkan I guess we are good now. At least I hope.

test/zimwriterfs-zimcreatorfs.cpp
kelson42: fixed.

veloman-yunkan (approved):
LGTM

PR #518, opened 2026-08-01
Classify arbitrary URI schemes as external

### LINKED ISSUE #517: zimcheck treats URIs without authority component as internal URLs if the scheme is not from a known limited set
zimcheck treats URIs of the form `scheme:path` (where path doesn't start with `//`) as internal URLs unless the scheme is from a known limited set. Currently supported schemes are `javascript`, `data`, `mailto`, `tel`, `sip`, `geo`, `xmpp`, `news`, & `urn`. That list is updated from time to time (see #205 & #329).

Note that only the `data` scheme receives special handling (as it is exempt from being reported as an external link). It is not clear why we have to recognize and discriminate other schemes instead of just marking all URIs of the form `scheme:...` (where `scheme` is not `data`) as external.

Originally reported in https://github.com/openzim/zim-tools/pull/511#issuecomment-5133709076

tomatotomata:
## Summary

- recognize valid URI schemes even when no authority component is present
- keep `data:` URIs exempt from external-link reporting
- preserve relative-path handling and add focused coverage for valid and invalid scheme forms

Closes #517

## Validation

- `git diff --check`
- Added coverage for empty, non-letter, and invalid-character scheme prefixes
- Native Meson, Ninja, compiler, and pkg-config tools are not installed in this Windows checkout, so the native test binary could not be built locally
- The repository CI provides the native build and test validation

## Scope

This change only adjusts URI classification and its focused tests. It does not change relative-path resolution or the handling of `data:` URIs.

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -569,6 +569,16 @@ std::string resolveLinkTarget(std::string url, const std::string& zimPath) {
 namespace
 {
 
+bool isAsciiAlpha(const char c)
+{
+    return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
+}
+
+bool isAsciiAlphaNumeric(const char c)
+{
+    return isAsciiAlpha(c) || (c >= '0' && c <= '9');
+}
+
 UriKind specialUriSchemeKind(const std::string& s)
 {
     static const std::map<std::string, UriKind> uriSchemes = {
@@ -623,7 +633,7 @@ static bool isReservedUrlChar(const char c)
 
 bool needsEscape(const char c, const bool encodeReserved)
 {
-  if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9'))
+  if (isAsciiAlphaNumeric(c))
     return false;
 
   if (isReservedUrlChar(c))
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -579,6 +579,18 @@ bool isAsciiAlphaNumeric(const char c)
     return isAsciiAlpha(c) || (c >= '0' && c <= '9');
 }
 
+bool isValidUriScheme(const std::string& scheme)
+{
+    if (scheme.empty() || !isAsciiAlpha(scheme.front())) {
+        return false;
+    }
+
+    return std::all_of(
+        scheme.begin() + 1,
+        scheme.end(),
+        [](const char c) { return isAsciiAlphaNumeric(c) || c == '+' || c == '-' || c == '.'; });
+}
+
 UriKind specialUriSchemeKind(const std::string& s)
 {
     static const std::map<std::string, UriKind> uriSchemes = {
@@ -594,7 +606,7 @@ UriKind specialUriSchemeKind(const std::string& s)
     };
 
     const auto it = uriSchemes.find(s);
-    return it != uriSchemes.end() ? it->second : UriKind::OTHER;
+    return it != uriSchemes.end() ? it->second : UriKind::GENERIC_URI;
 }
 
 
@@ -610,12 +622,16 @@ UriKind html_link::detectUriKind(const std::string& input_string)
             return UriKind::OTHER;
     }
 
+    const std::string raw_scheme = input_string.substr(0, k);
+    if (!isValidUriScheme(raw_scheme)) {
+        return UriKind::OTHER;
+    }
     if ( k + 2 < input_string.size()
          && input_string[k+1] == '/'
          && input_string[k+2] == '/' )
         return UriKind::GENERIC_URI;
 
-    const std::string scheme = asciitolower(input_string.substr(0, k));
+    const std::string scheme = asciitolower(raw_scheme);
     return specialUriSchemeKind(scheme);
 }
 
--- a/src/tools.h
+++ b/src/tools.h
@@ -71,7 +71,7 @@ enum class UriKind : int
     NEWS,           // news:comp.os.linux.announce
     URN,            // urn:nbn:de:bsz:24-digibib-bsz3530416370
 
-    GENERIC_URI,    // Generic URI with scheme and authority: <scheme>://.....
+    GENERIC_URI,    // Generic URI with a scheme: <scheme>:... or <scheme>://.....
     PROTOCOL_RELATIVE, // Protocol-relative URL: //<host>/<path>/<to>/<resource>
 
     OTHER           // not a valid URI (though it can be a valid relative
--- a/test/tools-test.cpp
+++ b/test/tools-test.cpp
@@ -202,17 +202,21 @@ TEST(tools, uriKind)
     EXPECT_EQ(UriKind::DATA, uriKind("data:text/plain;charset=UTF-8,data"));
     EXPECT_EQ(UriKind::DATA, uriKind("DATA:text/plain;charset=UTF-8,data"));
 
-    EXPECT_EQ(UriKind::OTHER, uriKind("http:example.com"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("http:/example.com"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("http:example.com"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("http:/example.com"));
     EXPECT_EQ(UriKind::OTHER, uriKind("git@github.com:openzim/zim-tools.git"));
     EXPECT_EQ(UriKind::OTHER, uriKind("/redirect?url=http://example.com"));
     EXPECT_EQ(UriKind::OTHER, uriKind("redirect?url=http://example.com"));
     EXPECT_EQ(UriKind::OTHER, uriKind("auth.php#returnurl=https://example.com"));
     EXPECT_EQ(UriKind::OTHER, uriKind("/api/v1/http://example.com"));
     EXPECT_EQ(UriKind::OTHER, uriKind("img/file:///etc/passwd"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("ftp:/download.kiwix.org/zim/"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("sendmailto:someone@example.com"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("intel:+0123456789"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("ftp:/download.kiwix.org/zim/"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("sendmailto:someone@example.com"));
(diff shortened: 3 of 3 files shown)


kelson42:
@tomatotomata Can you please format the PR description properly?

tomatotomata:
I reformatted the PR body with real Markdown line breaks and explicit Summary, Validation, and Scope sections. I also added focused tests for an empty scheme, a non-letter prefix, and an invalid underscore so the new rejection path is covered. The update is on commit `2552f3c`; the native suite remains CI-only from this Windows checkout.

kelson42:
@tomatotomata Thank you, can you please also rebase (no merge of main) on main head (and resolve conflict)?

tomatotomata:
I rebased the PR onto the current `main` without merging it and resolved the only conflict in `src/tools.cpp` by keeping both the existing Unicode includes and the new character-classification include. The rebased head is `dd2503a` and is force-pushed to the PR branch.

`git diff --check` passes. The native Meson test suite remains CI-only from this Windows checkout, so the refreshed Codecov and CI results are the next useful signal. Let me know what you think.

tomatotomata:
I pushed `5e36806` with the follow-up cleanup: the unused `<cctype>` include is gone, URI schemes are validated from the raw substring before the authority-form branch, and lowercasing is now limited to special-scheme lookup. The ASCII-only validator and its UTF-8 and mixed-character regressions remain in place.

`git diff --check` passes. The latest fork CI and Packages runs are waiting on GitHub Actions approval, so I am not claiming a native suite result from this head. Codecov on the previous head reached 71.4% but still reported four partial lines; this head is queued for a fresh report. Let me know what you think.

kelson42:
@tomatotomata Copilot is over now, maybe you can squash everything in one or teo comments before human code review by @veloman-yunkan starts?

src/tools.h
```diff
@@ -71,6 +71,7 @@ enum class UriKind : int
     NEWS,           // news:comp.os.linux.announce
     URN,            // urn:nbn:de:bsz:24-digibib-bsz3530416370
 
+    URI_WITH_SCHEME, // Generic URI without an authority component: <scheme>:...
```
veloman-yunkan: I don't think that we need a new enumeration value.

I am rather inclined to reduce the list of members of `UriKind` to the smallest set justified by the current implementation (which would be `DATA`, `GENERIC_URI` and `OTHER`) but I don't feel comfortable dropping `JAVASCRIPT` and `PROTOCOL_RELATIVE`.

For the time being let's just redefine the cases covered by `GENERIC_URI` (to include URIs both with or without authority).

test/tools-test.cpp
```diff
@@ -202,17 +202,27 @@ TEST(tools, uriKind)
...
-    EXPECT_EQ(UriKind::OTHER, uriKind("ftp:/download.kiwix.org/zim/"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("sendmailto:someone@example.com"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("intel:+0123456789"));
+    EXPECT_EQ(UriKind::URI_WITH_SCHEME, uriKind("ftp:/download.kiwix.org/zim/"));
+    EXPECT_EQ(UriKind::URI_WITH_SCHEME, uriKind("sendmailto:someone@example.com"));
+    EXPECT_EQ(UriKind::URI_WITH_SCHEME, uriKind("intel:+0123456789"));
+    EXPECT_EQ(UriKind::URI_WITH_SCHEME, uriKind("custom-scheme:opaque-value"));
+    EXPECT_EQ(UriKind::URI_WITH_SCHEME, uriKind("My2+scheme.value:opaque-value"));
+    EXPECT_TRUE(html_link(html_link::HREF, "custom-scheme:opaque-value").isExternalUrl());
+    EXPECT_FALSE(html_link(html_link::HREF, "data:text/plain,payload").isExternalUrl());
```
veloman-yunkan: These two test points look out of place. Better add a new unit test `tools.linkClassification` with helper macros `EXPECT_INTERNAL_LINK` and `EXPECT_EXTERNAL_LINK` encapsulating the `html_link` (testing it for both `HREF` and `SRC`) so that the checks read:

```c++
EXPECT_INTERNAL_LINK("data:text/plain,payload");
EXPECT_EXTERNAL_LINK("custom-scheme:opaque-value");
```

Of course there should be more test points in that new unit test.

src/tools.cpp
```diff
@@ -569,6 +569,28 @@ std::string resolveLinkTarget(std::string url, const std::string& zimPath) {
...
 
+bool isAsciiAlpha(const char c)
+{
+    return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
+}
+
+bool isAsciiAlphaNumeric(const char c)
+{
+    return isAsciiAlpha(c) || (c >= '0' && c <= '9');
+}
```
veloman-yunkan, line 580: Having introduced this function you can immediately take advantage of it in `needsEscape()`

veloman-yunkan (changes requested):
Please structure your PR as two commits:

1. A small refactoring step: introduce `isAsciiAlpha()` and `isAsciiAlphaNumeric()` by extracting them from `needsEscape()`
2. The rest of changes in one commit

tomatotomata:
I reworked the branch into the two-commit structure you suggested:

1. `1f4eba1` extracts the shared ASCII alpha/alphanumeric checks from `needsEscape()` and uses them for URI-scheme validation.
2. `d674dba` keeps `UriKind` to the existing three generic categories, treats valid scheme-only URIs as `GENERIC_URI`, updates the `HREF`/`SRC` link-classification coverage, and adds the invalid UTF-8 scheme cases.

I also moved the link tests into `tools.linkClassification` with helper macros so both attributes are exercised. I could not run the CMake test binary because CMake is not installed in this Windows environment; `git diff --check` passes. Let me know what you think.

src/tools.cpp
```diff
@@ -621,9 +610,19 @@ static bool isReservedUrlChar(const char c)
...
+
```
veloman-yunkan: Place these near the top of the file when extracting them so that you don't need a separate declaration when using them in the next commit

src/tools.cpp
```diff
@@ -403,10 +403,10 @@ std::vector<html_link> generic_getLinks(const std::string& page)
         }
 
         if (strncmp(p, " href", 5) == 0) {
-            attr = "href";
+            attr = html_link::HREF;
```
veloman-yunkan, line 423: Why do these (and a few other changes from another PR) show up here? Please rebase your branch on main.

test/tools-test.cpp
```diff
@@ -222,6 +226,29 @@ TEST(tools, uriKind)
...
+
+#define EXPECT_EXTERNAL_LINK(url) \
+    do { \
+        EXPECT_TRUE(html_link(html_link::HREF, (url)).isExternalUrl()); \
+        EXPECT_TRUE(html_link(html_link::SRC, (url)).isExternalUrl()); \
+    } while (false)
+
+TEST(tools, linkClassification)
+{
+    EXPECT_INTERNAL_LINK("data:text/plain,payload");
```
veloman-yunkan, line 243: Please add a few more test points for internal links

test/tools-test.cpp
```diff
@@ -202,17 +202,21 @@ TEST(tools, uriKind)
...
-    EXPECT_EQ(UriKind::OTHER, uriKind("ftp:/download.kiwix.org/zim/"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("sendmailto:someone@example.com"));
-    EXPECT_EQ(UriKind::OTHER, uriKind("intel:+0123456789"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("ftp:/download.kiwix.org/zim/"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("sendmailto:someone@example.com"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("intel:+0123456789"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("custom-scheme:opaque-value"));
+    EXPECT_EQ(UriKind::GENERIC_URI, uriKind("My2+scheme.value:opaque-value"));
+    EXPECT_EQ(UriKind::OTHER, uriKind(std::string("\xC3\xA9") + "custom:value"));
+    EXPECT_EQ(UriKind::OTHER, uriKind(std::string("custom") + "\xC3\xA9:value"));
```
veloman-yunkan: Why don't you write simply `"\xC3\xA9custom:value"` and `"custom\xC3\xA9:value"`?

veloman-yunkan (changes requested):

kelson42:
@tomatotomata Hi, sorry to be a bit pushy, but this important PR is the last one in our milestone and I would like - at the same time - to have it merged AND release the zim-tools `3.8.0` worse case tomorrow evening UTC timezone. Do you think you could have a look, otherwise I @veloman-yunkan would fix the last few details himself so we can move forward?

veloman-yunkan:
@tomatotomata There have been no published updates since my last review. The PR status reads "This branch has conflicts that must be resolved".

tomatotomata:
I rebased the branch onto the current upstream `main` and resolved the conflict in `src/tools.cpp`. The refreshed head is `c1c3d03`.

The Windows Meson build now completes all eight targets. `metadata-test` passes all 12 tests. `zimcheck-test` still exits with Windows access violation `0xc0000005`, so I am leaving that result visible rather than calling the full suite green.

The branch is force-pushed and ready for the refreshed checks. Let me know what you think.

tomatotomata:
With CMake installed, I reran the native build path. The CMake-configured libzim dependency builds successfully with `cmake --build ... --config Debug --parallel 2`; `ctest` reports no tests registered in that dependency project. zim-tools itself uses Meson, so I also ran its tests on `c1c3d03`: `metadata-test` passes all 12 tests, while `zimcheck-test` still exits with Windows `0xc0000005`. I am leaving the full suite marked partial rather than green. Let me know what you think.

src/tools.cpp
veloman-yunkan: This comment was not addressed. I took the liberty of doing it myself.

test/tools-test.cpp
veloman-yunkan: This comment was not addressed. I took the liberty of doing it myself.

test/tools-test.cpp
veloman-yunkan: I see now why `"\xC3\xA9custom:value"` doesn't work. According to https://cppreference.com/cpp/language/string_literal:

> If a valid hexadecimal digit follows a hexadecimal escape sequence in a string literal, it would fail to compile as an invalid escape sequence. String concatenation can be used as a workaround.

veloman-yunkan:
The comments from [my last review](https://github.com/openzim/zim-tools/pull/518#pullrequestreview-4897849347) had not been addressed so I followed up on them myself.

veloman-yunkan (approved):

PR #519, opened 2026-08-01
Treat unknown URI schemes as external

### LINKED ISSUE #517: zimcheck treats URIs without authority component as internal URLs if the scheme is not from a known limited set

zzz-web-zz1:
## Summary

- validate URI schemes using the RFC 3986 scheme syntax
- classify unknown but valid schemes as generic external URIs
- preserve special handling for known schemes such as `data:`
- add regression coverage for valid and invalid custom schemes

## Why

`detectUriKind()` previously treated a URI as external only when it used a known scheme or contained an authority component (`://`). Valid URIs with an unknown scheme and no authority component were therefore treated as internal links by `zimcheck`.

This change makes any syntactically valid explicit scheme external while keeping relative links and malformed schemes internal.

## Validation

- `git diff --check`
- focused regression vectors validated locally
- full C++ build and test suite deferred to GitHub Actions because the local environment does not contain the project's native build dependencies

Fixes #517

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -576,6 +576,21 @@ UriKind specialUriSchemeKind(const std::string& s)
     return it != uriSchemes.end() ? it->second : UriKind::OTHER;
 }
 
+bool isValidUriScheme(const std::string& s)
+{
+    const auto isAlpha = [](const char c) {
+        return ('a' <= c && c <= 'z') || ('A' <= c && c <= 'Z');
+    };
+
+    if (s.empty() || !isAlpha(s.front()))
+        return false;
+
+    return std::all_of(s.begin() + 1, s.end(), [&isAlpha](const char c) {
+        return isAlpha(c) || ('0' <= c && c <= '9')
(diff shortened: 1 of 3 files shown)


kelson42:
@zzz-web-zz1 Hi, is this PR ready for a review? Would you please add the issue it fixes in the PR description?

kelson42:
Closed in favour of #518

PR #520, opened 2026-08-05
Fixed zimsplit's --size option

### LINKED ISSUE #434: zimsplit (zim-tools_win-x86_64-3.5.0) can not split big files (100gb) to bigger slices than ca. 2,1Gbyte
I tried to split wikipedia_en_all_maxi_2024-01.zim with 102 GB (109.885.670.576 Bytes) to 25gb slices, and got "stol argument out of range".

-------
zim-tools 3.5.0

libzim 9.2.3

    libzstd 1.5.5
    liblzma 5.2.11
    libxapian 1.4.23

veloman-yunkan:
Fixes #434 

The value of the `--size` option is now parsed as a 64-bit value (regardless of the bit-width of the `long int` type of the compiler used). This change also fixes a bug where negative values for the `--size` option were not rejected but were silently converted to a large positive value.  

Since the bug was about parsing of a numeric value exceeding the 32-bit range I only tested that the new parsing code works (didn't run `zimsplit` on a huge ZIM file using chunk size in excess of 2GiB).

### DIFF
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -92,7 +92,7 @@ class ZimSplitter
 
     void close_file() {
         if (currentPartSize > maxPartSize) {
-           std::cout << "WARNING: Part " << part_name << " is bigger that max part size."
+           std::cout << "WARNING: Part " << part_name << " is bigger than max part size."
             << " (" << currentPartSize << ">" << maxPartSize << ")" << std::endl;
         }
         ofile.close();
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -30,7 +30,18 @@
 
 #define BUFFER_SIZE 4096
 
-#define DEFAULT_PART_SIZE 2147483648
+const zim::size_type DEFAULT_PART_SIZE = 2147483648;
+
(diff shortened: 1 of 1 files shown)


kelson42:
@veloman-yunkan Not sure if this PR is ready for review, but I made:
```
$ zimsplit --size=20GB /home/kelson/Downloads/wikipedia_fr_all_maxi_2026-05.zim 
invalid uint64_t value: 20GB
```

Maybe the problem is the value given, the usage() is not clear about this.

veloman-yunkan:
@kelson42 That was never supposed to work - you have to provide a numeric value in bytes. The old version gave a slightly better error message (20GB contains non-numeric characters).

kelson42:
> @kelson42 That was never supposed to work - you have to provide a numeric value in bytes. 

I have created https://github.com/openzim/zim-tools/issues/524 to improve this.

kelson42:
@veloman-yunkan It seems to work for me, should I merge the PR?

veloman-yunkan:
@kelson42 Let me toss a coin... It came heads. So let's merge the PR.

PR #530, opened 2026-08-10
Validation of MIME type for reserved metadata

### LINKED ISSUE #474: zimcheck: check known/expected metadata mimetype
As mentioned by @rgaudin  in https://github.com/openzim/mwoffliner/issues/2470, it seems that all known/expected metadata (but illustrations) should have `text/plain` or `text/plain;charset=UTF-8` mimetype.

It looks like zimcheck is not checking this at all.

Probably worth to add. Not sure if it should be an error or a warning.

veloman-yunkan:
Fixes #474

### DIFF
--- a/src/metadata.cpp
+++ b/src/metadata.cpp
@@ -37,6 +37,8 @@ namespace
 const bool MANDATORY = true;
 const bool OPTIONAL  = false;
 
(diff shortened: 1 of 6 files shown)


kelson42 (approved):
LGTM

PR #532, opened 2026-08-13
Froze the version of meson in CI for Win & macOS

### LINKED ISSUE #531: zim-tools build fails with a linker error (actually a warning) when using meson 1.12.0
See https://github.com/kiwix/kiwix-build/actions/runs/31449591905

veloman-yunkan:
Fixes #531

Windows and macOS CI jobs used the latest available version of meson. A recent upgrade of it from v1.11.2 to v1.12.0 has resulted in the failure of the Windows build (caused by a linker warning now being treated as error). So sticking to the last version which has worked.

Note that currently in all of our Linux builder images v1.6.1 of meson is installed, while flatpak provides v1.3.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
           brew install gcovr || brew link --overwrite python # ninja
 
       - name: Install Python modules
-        run: pip3 install meson pytest
+        run: pip3 install meson==1.11.2 pytest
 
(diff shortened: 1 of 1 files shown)


kelson42:
You want to open a separate issue for the warning to fix?

veloman-yunkan:
No, I don't. That warning has been there for quite some time and can hardly cause any trouble.

PR #537, opened 2026-08-18
Support size units in zimsplit

### LINKED ISSUE #524: zimsplit --size command line should be given using a nice format
Today we can only give the number of bytes for each chunk, but it would be nice to be able to just write something like 2GB.

Once fixed the `usage()` and the man file should be updated.

Nidhi-Gahlawat:
## Summary

- Preserve unitless `--size` values as bytes.
- Support decimal units (`KB`, `MB`, ..., `EB`) and binary units (`KiB`, `MiB`, ..., `EiB`).
- Match units case-insensitively and ignore whitespace around the size and before its unit.
- Reject zero, malformed values, and values that overflow `uint64_t`.
- Require the requested part size to be smaller than the input ZIM file before creating output files.
- Update CLI help, the zimsplit man page, and focused parser/validation tests.

## Commit structure

1. `Reject oversized zimsplit part sizes`
2. `Support size units in zimsplit`

## Testing

- `meson test -C build --print-errorlogs` (all 5 test targets pass)
- Focused `zimsplit-test` under AddressSanitizer and UndefinedBehaviorSanitizer
- Coverage build: `src/zimsplit_size.cpp` has 100% line coverage
- Manual CLI check with `--size="    1    GB   "` reaches source-size validation

Closes #524

### DIFF
--- a/src/zimsplit_size.cpp
+++ b/src/zimsplit_size.cpp
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2026 Nidhi Gahlawat.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#include "zimsplit_size.h"
+
+#include <stdexcept>
+
+namespace
+{
+
+constexpr auto partSizeTooLargeError
+    = "part size must be smaller than the input ZIM file";
+
+}  // namespace
+
+void validatePartSize(uint64_t partSize, uint64_t archiveSize)
+{
+  if (partSize >= archiveSize) {
+    throw std::invalid_argument(partSizeTooLargeError);
+  }
+}
--- a/src/zimsplit_size.cpp
+++ b/src/zimsplit_size.cpp
@@ -18,8 +18,14 @@
  */
 
 #include "zimsplit_size.h"
+#include "tools.h"
 
+#include <array>
+#include <charconv>
+#include <limits>
 #include <stdexcept>
+#include <string>
+#include <utility>
 
 namespace
 {
@@ -27,7 +33,80 @@ namespace
 constexpr auto partSizeTooLargeError
     = "part size must be smaller than the input ZIM file";
 
+void trimAsciiWhitespace(std::string& value)
+{
+  constexpr auto whitespace = " \t\n\r\f\v";
+  const auto first = value.find_first_not_of(whitespace);
+  if (first == std::string::npos) {
+    value.clear();
+    return;
+  }
+
+  const auto last = value.find_last_not_of(whitespace);
+  value = value.substr(first, last - first + 1);
+}
+
+uint64_t getUnitMultiplier(std::string unit)
+{
+  static const std::array<std::pair<std::string, uint64_t>, 14>
+      multipliers{{
+          {"", 1},
+          {"B", 1},
+          {"KB", 1000ULL},
+          {"MB", 1000ULL * 1000},
+          {"GB", 1000ULL * 1000 * 1000},
+          {"TB", 1000ULL * 1000 * 1000 * 1000},
+          {"PB", 1000ULL * 1000 * 1000 * 1000 * 1000},
+          {"EB", 1000ULL * 1000 * 1000 * 1000 * 1000 * 1000},
+          {"KiB", 1ULL << 10},
+          {"MiB", 1ULL << 20},
+          {"GiB", 1ULL << 30},
+          {"TiB", 1ULL << 40},
+          {"PiB", 1ULL << 50},
+          {"EiB", 1ULL << 60},
+      }};
+
+  const auto normalizedUnit = asciitolower(unit);
+  for (const auto& multiplier : multipliers) {
+    if (asciitolower(multiplier.first) == normalizedUnit) {
+      return multiplier.second;
+    }
+  }
+
+  throw std::invalid_argument("invalid size unit: " + unit);
+}
+
 }  // namespace
+
+uint64_t parseByteSize(std::string value)
+{
+  trimAsciiWhitespace(value);
+  if (value.empty()) {
+    throw std::invalid_argument("invalid size value: empty input");
+  }
+
+  uint64_t size = 0;
+  const char* const end = value.data() + value.size();
+  const auto parseResult = std::from_chars(value.data(), end, size);
+  if (parseResult.ec == std::errc::result_out_of_range) {
+    throw std::invalid_argument(partSizeTooLargeError);
+  }
+  if (parseResult.ec != std::errc() || parseResult.ptr == value.data()) {
+    throw std::invalid_argument("invalid size value: " + value);
+  }
+
+  std::string unit(parseResult.ptr, end);
+  trimAsciiWhitespace(unit);
+  const auto multiplier = getUnitMultiplier(unit);
+  if (size == 0) {
+    throw std::invalid_argument("part size must be positive: " + value);
+  }
+  if (size > std::numeric_limits<uint64_t>::max() / multiplier) {
+    throw std::invalid_argument(partSizeTooLargeError);
+  }
+
+  return size * multiplier;
+}
 
 void validatePartSize(uint64_t partSize, uint64_t archiveSize)
 {
--- a/src/zimsplit_size.h
+++ b/src/zimsplit_size.h
@@ -21,7 +21,9 @@
 #define OPENZIM_ZIMSPLIT_SIZE_H
 
 #include <cstdint>
+#include <string>
 
+uint64_t parseByteSize(std::string value);
 void validatePartSize(uint64_t partSize, uint64_t archiveSize);
 
 #endif
--- a/src/zimsplit_size.h
+++ b/src/zimsplit_size.h
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2026 Nidhi Gahlawat.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU  General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#ifndef OPENZIM_ZIMSPLIT_SIZE_H
+#define OPENZIM_ZIMSPLIT_SIZE_H
+
+#include <cstdint>
+
+void validatePartSize(uint64_t partSize, uint64_t archiveSize);
+
+#endif
--- a/test/zimsplit-test.cpp
+++ b/test/zimsplit-test.cpp
@@ -1,0 +1,31 @@
+#include <stdexcept>
+
+#include "../src/zimsplit_size.h"
+#include "gtest/gtest.h"
+
+namespace
+{
+
+#define EXPECT_INVALID_ARG_ERROR(statement, expectedError)                      \
+  do {                                                                         \
+    try {                                                                      \
+      statement;                                                               \
+      FAIL() << "Expected std::invalid_argument";                              \
+    } catch (const std::invalid_argument& error) {                              \
+      EXPECT_STREQ(error.what(), expectedError);                                \
+    }                                                                          \
+  } while (false)
+
+constexpr auto TOO_LARGE_PART_SIZE_ERROR
+    = "part size must be smaller than the input ZIM file";
+
+}  // namespace
+
+TEST(ZimSplitSize, ValidatesPartSizeAgainstArchive)
+{
+  EXPECT_NO_THROW(validatePartSize(99, 100));
+  EXPECT_INVALID_ARG_ERROR(validatePartSize(100, 100),
+                           TOO_LARGE_PART_SIZE_ERROR);
+  EXPECT_INVALID_ARG_ERROR(validatePartSize(101, 100),
+                           TOO_LARGE_PART_SIZE_ERROR);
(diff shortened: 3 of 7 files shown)


src/zimsplit_size.cpp
```diff
@@ -0,0 +1,67 @@
+#include "zimsplit_size.h"
+
+#include <array>
```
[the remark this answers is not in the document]
Nidhi-Gahlawat: Added the repository-standard GPL header in f9d3116. The focused zimsplit parser tests still pass.

src/zimsplit_size.h
```diff
@@ -0,0 +1,9 @@
+#ifndef OPENZIM_ZIMSPLIT_SIZE_H
+#define OPENZIM_ZIMSPLIT_SIZE_H
+
```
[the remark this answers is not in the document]
Nidhi-Gahlawat: Added the repository-standard GPL header in f9d3116. The focused zimsplit parser tests still pass.

src/zimsplit_size.cpp
```diff
@@ -0,0 +1,67 @@
...
+          {"TB", 1000ULL * 1000 * 1000 * 1000},
+          {"PB", 1000ULL * 1000 * 1000 * 1000 * 1000},
+          {"EB", 1000ULL * 1000 * 1000 * 1000 * 1000 * 1000},
+          {"KiB", 1ULL << 10},
+          {"MiB", 1ULL << 20},
+          {"GiB", 1ULL << 30},
+          {"TiB", 1ULL << 40},
+          {"PiB", 1ULL << 50},
+          {"EiB", 1ULL << 60},
+      }};
```
veloman-yunkan, line 67: Do we really need to support SI units above GiB?

It may be more practical to support sizes for typical storage/archival media (such as "CD", "DVD", "BD") than SI units (such as PiB) that are hardly going to be used for splitting ZIM files in the foreseeable future.

Also I think that units can be made case-insensitive.

src/zimsplit_size.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+}  // namespace
+
+uint64_t parseByteSize(std::string_view value)
+{
+  const auto unitOffset = value.find_first_not_of("0123456789");
+  const auto number = value.substr(0, unitOffset);
+  const auto unit = unitOffset == std::string_view::npos
+                        ? std::string_view()
+                        : value.substr(unitOffset);
```
veloman-yunkan: Why don't you rely on `std::from_chars()` to find the end of the numeric value (`parseResult.ptr`)?

src/zimsplit_size.cpp
```diff
@@ -0,0 +1,67 @@
...
+      = std::from_chars(number.data(), number.data() + number.size(), size);
+  if (number.empty() || parseResult.ec != std::errc()
+      || parseResult.ptr != number.data() + number.size()) {
+    throw std::invalid_argument("invalid size value: " + std::string(value));
+  }
+
+  const auto multiplier = getUnitMultiplier(unit);
+  if (size == 0 || size > std::numeric_limits<uint64_t>::max() / multiplier) {
+    throw std::invalid_argument("size must be positive and fit in uint64_t: "
+                                + std::string(value));
```
veloman-yunkan: This makes sense for a general-purpose function `parseByteSize()`, but not for a zim-split specific one.

I think that the application should be foolproof and reject unreasonably large chunk sizes right away.

veloman-yunkan (commented):
Code looks good but I have some concerns related to UX.

src/zimsplit_size.cpp
kelson42: sorry, but why? If the chunk is larger than the ZIM file, then it will do nothing! What is wrong with that?

veloman-yunkan: I checked that if `zimsplit` is called with a chunk size exceeding the size of the ZIM file, it still creates a copy of the file with the `zimaa` extension. Therefore if the user for whatever reason provides an unreasonably high value for the chunk size (making it pointless to try to split the provided ZIM file into pieces of that size), `zimsplit` will run for a few minutes on large files wasting the user's time. Therefore I think that chunk size should be bounded from above by the size of the ZIM file.

BTW, during testing I found out that the option `--force` of `zimsplit` is either obsolete or there is a bug. Running `zimsplit` with a chunk size below the smallest possible part size for a given ZIM file completes successfully with warnings issued for parts that had to exceed the user-specified chunk-size.

src/zimsplit_size.cpp
Nidhi-Gahlawat: Makes sense, the parser now uses parseResult.ptr to identify the unit suffix. Thanks!

src/zimsplit_size.cpp
Nidhi-Gahlawat: Thanks. I made unit matching ASCII case-insensitive, added mixed-case tests, and updated the CLI help and man page.
I kept the decimal and binary unit range currently documented because issue #524 explicitly requested both families. I did not add CD, DVD, or BD aliases because their capacities are ambiguous across media variants; I think those would be better specified in a separate issue.

src/zimsplit_size.cpp
Nidhi-Gahlawat: Added a check, now `zimsplit` rejects a requested part size greater than or equal to the input ZIM size before creating any output files. The `--force` behavior predates this parser change, so I suggest tracking it separately.

src/zimsplit_size.cpp
```diff
@@ -29,6 +29,27 @@
...
+
+  for (std::string_view::size_type index = 0; index < lhs.size(); ++index) {
+    if (toLowerAscii(lhs[index]) != toLowerAscii(rhs[index])) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
```
veloman-yunkan: This function is not justified. Just keep the unit names in your `multipliers` collection in `getUnitMultiplier()` in lower case. Then only the parameter `unit` of `getUnitMultiplier()` will need to be converted to lower-case during run-time.

src/zimsplit_size.cpp
veloman-yunkan: 1. For size values that don't "fit in uint64_t" `zimsplit` will report this error, whereas, IMHO, it better report "part size must be smaller than the input ZIM file".

2. Do we really want to allow small part sizes? The current check will happily accept 1 byte parts. I know that we better have a separate ticket about that change. @kelson42 Will you please open new tickets based on those suggestions found in the review comments of this PR which you agree with?

veloman-yunkan (changes requested):

src/zimsplit_size.cpp
Nidhi-Gahlawat: Values exceeding the supported range now report the suggested error message. Thanks!

src/zimsplit_size.cpp
Nidhi-Gahlawat: Makes sense, unit names are now stored in lowercase, and the input unit is normalized once before lookup.

src/zimsplit_size.cpp
```diff
@@ -29,49 +29,38 @@
...
+          {"pib", 1ULL << 50},
+          {"eib", 1ULL << 60},
       }};
 
+  std::string normalizedUnit(unit);
+  for (auto& character : normalizedUnit) {
+    if (character >= 'A' && character <= 'Z') {
+      character = static_cast<char>(character + ('a' - 'A'));
+    }
+  }
```
veloman-yunkan: Use `asciitolower()` from `src/tools.h`. I don't know what I was thinking yesterday, but I looked at `equalsIgnoreAsciiCase()` inattentively (since I was confident that it should not have been introduced no matter how it was implemented) and was under the impression that `toLowerAscii()` referred to the function that we already had (`asciitolower()`). Given that we can convert a string to lower-case easily, keeping the unit names in lower-case somewhat hurts readability and this is not a place where performance matters. Thus I don't mind if you restore the unit-names and normalize them on the fly (right in the condition of the `if`). I leave it up to you to decide on this matter.

test/zimsplit-test.cpp
```diff
@@ -5,6 +5,21 @@
...
+
+void expectInvalidArgument(std::string_view value, const char* message)
+{
+  try {
+    parseByteSize(value);
+    FAIL() << "Expected std::invalid_argument";
+  } catch (const std::invalid_argument& error) {
+    EXPECT_STREQ(error.what(), message);
+  }
+}
```
veloman-yunkan: Please rename to `EXPECT_INVALID_ARG_ERROR()` and convert to a macro (so that in case of failure the reported line is in the unit test function.

test/zimsplit-test.cpp
veloman-yunkan: Shouldn't these checks be replaced?

test/zimsplit-test.cpp
```diff
@@ -69,9 +84,10 @@ TEST(ZimSplitSize, RejectsMalformedValues)
 
 TEST(ZimSplitSize, RejectsOverflow)
 {
-  EXPECT_THROW(parseByteSize("18446744073709551616"), std::invalid_argument);
-  EXPECT_THROW(parseByteSize("19EB"), std::invalid_argument);
-  EXPECT_THROW(parseByteSize("16EiB"), std::invalid_argument);
+  constexpr auto error = "part size must be smaller than the input ZIM file";
```
veloman-yunkan: Rename to TOO_LARGE_PART_SIZE_ERROR, move to namespace scope and reuse in the `ZimSplitSize.ValidatesPartSizeAgainstArchive` unit-test

veloman-yunkan (changes requested):
I propose to structure this PR as two commit:

1. Reject part sizes exceeding the ZIM file size (introduce `validatePartSize()` and its unit-test in new test file `zimsplit-test.cpp`.
2. Rest of the changes (support for size units)

src/zimsplit_size.cpp
Nidhi-Gahlawat: Addressed in `4884c83`. `getUnitMultiplier()` now uses the existing `asciitolower()` helper and keeps the unit names in their canonical form.

test/zimsplit-test.cpp
Nidhi-Gahlawat: Addressed in `4884c83`. The helper is now an `EXPECT_INVALID_ARG_ERROR()` macro, so failures are reported at the calling test line.

test/zimsplit-test.cpp
Nidhi-Gahlawat: Addressed this,  the remaining rejection checks now use `EXPECT_INVALID_ARG_ERROR()` and verify the exact error messages.

test/zimsplit-test.cpp
Nidhi-Gahlawat: Addressed in `4884c83`. `TOO_LARGE_PART_SIZE_ERROR` is now at namespace scope and reused by both overflow and archive-size validation tests.

src/zimsplit_size.cpp
```diff
@@ -18,17 +18,82 @@
...
+          {"MiB", 1ULL << 20},
+          {"GiB", 1ULL << 30},
+          {"TiB", 1ULL << 40},
+          {"PiB", 1ULL << 50},
+          {"EiB", 1ULL << 60},
+      }};
+
+  for (const auto& multiplier : multipliers) {
+    if (asciitolower(std::string(multiplier.first))
+        == asciitolower(std::string(unit))) {
```
veloman-yunkan: Readability of this expression suffers from the fact that the implementation of this absolutely non-performance-critical functionality uses `string_view` instead of `string`. Let's not plant a seed of all evil in this PR :)

test/zimsplit-test.cpp
```diff
@@ -21,6 +23,84 @@ constexpr auto TOO_LARGE_PART_SIZE_ERROR
...
+  EXPECT_INVALID_ARG_ERROR(parseByteSize(std::string_view()),
+                           "invalid size value: empty input");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize("-1"), "invalid size value: -1");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize("+1"), "invalid size value: +1");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize("1.5GB"),
+                           "invalid size unit: .5GB");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize("1 GB"), "invalid size unit:  GB");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize(" 1GB"),
+                           "invalid size value:  1GB");
+  EXPECT_INVALID_ARG_ERROR(parseByteSize("1GB "), "invalid size unit: GB ");
```
veloman-yunkan: It's good that you took advantage of `EXPECT_INVALID_ARG_ERROR()` here as well. As you can see the error message shown to the user may be confusing (because whitespace are usually invisible).

One solution is to make whitespace visible via usage of quotes. Another solution is to strip leading and trailing whitespace from the argument of `parseByteSize()` as well as from the unit part of it (so that " 1 GB", "1GB ", "    1    GB   " are all accepted as valid inputs for the `--size` option of `zimsplit`. I consider the second approach more user-oriented.

veloman-yunkan (changes requested):
The PR description must be updated.

test/zimsplit-test.cpp
Nidhi-Gahlawat: Good point, I went with the second approach. The parser now accepts whitespace around the value and before the unit, and I added tests for those examples. I also updated the help text and man page to mention it.

src/zimsplit_size.cpp
Nidhi-Gahlawat: Fair enough, I’ve switched it to std::string, which makes the code much easier to read.

src/zimsplit_size.cpp
```diff
@@ -18,17 +18,96 @@
...
+          {"gb", 1000ULL * 1000 * 1000},
+          {"tb", 1000ULL * 1000 * 1000 * 1000},
+          {"pb", 1000ULL * 1000 * 1000 * 1000 * 1000},
+          {"eb", 1000ULL * 1000 * 1000 * 1000 * 1000 * 1000},
+          {"kib", 1ULL << 10},
+          {"mib", 1ULL << 20},
+          {"gib", 1ULL << 30},
+          {"tib", 1ULL << 40},
+          {"pib", 1ULL << 50},
+          {"eib", 1ULL << 60},
```
veloman-yunkan: I wonder why you chose to revert unit names to lower case.

veloman-yunkan (commented):
LGTM. We can merge either way. But I'd like to give you a chance to correct any unintentional changes from the previous iteration.

src/zimsplit_size.cpp
Nidhi-Gahlawat: I changed them to lowercase so the input only needed to be normalized once before lookup. But I agree that keeping the canonical unit names makes the table clearer, so I’ve restored them. Thanks!

veloman-yunkan (approved):

kelson42:
@Nidhi-Gahlawat Thank you very much for the good work!

veloman-yunkan:
@kelson42 New tickets can be opened based on the following discussions above:

- https://github.com/openzim/zim-tools/pull/537#discussion_r3831469793
- https://github.com/openzim/zim-tools/pull/537#discussion_r3839203702

src/zimsplit_size.cpp
kelson42: @Nidhi-Gahlawat Issue open at https://github.com/openzim/zim-tools/issues/549

src/zimsplit_size.cpp
kelson42: @Nidhi-Gahlawat Issue open at https://github.com/openzim/zim-tools/issues/550

PR #540, opened 2026-08-26
Check MIME types against filename extensions

### LINKED ISSUE #187: Check if filename extensions are coherent with mime-type
* If not extension at all, don't check.
* If wrong generate a warning

Nidhi-Gahlawat:
Fixes #187.

- Add `-T`/`--mime_type` to check filename-extension and MIME-type compatibility.
- Report incompatible combinations as warnings.
- Skip entries without extensions and unknown extensions.
- Accept case differences, MIME parameters, and valid alternative MIME types.
- Add CLI, text-output, JSON-output, and compatibility tests.
- Replace the stale `-E`/`--mime` man-page entry.

## Testing
- Built all targets
- Manually checked a generated ZIM archive using `zimcheck -T ../mimecheck.zim`


kelson42:
@Nidhi-Gahlawat Thank you very much for the very quick reaction and PR. One thing which seems strange to me is that we don't have this kind of table associated file extensions and mimeTypes already somewhere in our code... or in libmagic.

Nidhi-Gahlawat:
> @Nidhi-Gahlawat Thank you very much for the very quick reaction and PR. One thing which seems strange to me is that we don't have this kind of table associated file extensions and mimeTypes already somewhere in our code... or in libmagic.

Good point. There is already `extMimeTypes` in`src/zimwriterfs/tools.cpp`, and `zimwriterfs` uses libmagic as a fallback.

I did not reuse `extMimeTypes` directly because it selects one preferred MIME type when creating an entry, while validation needs to accept multiple valid MIME types for some extensions. For example, `text/javascript` is the current MIME type for `.js`, but existing archives may legitimately use the deprecated `application/javascript` alias.

The existing table is also private to the optional `zimwriterfs` target. libmagic identifies MIME types from file contents, but does not provide an extension-to-compatible-MIME-types relation.

Alternatively, I can move the compatibility information into shared tools while keeping separate APIs for selecting a preferred MIME type and validating compatible MIME types.

kelson42:
> > @Nidhi-Gahlawat Thank you very much for the very quick reaction and PR. One thing which seems strange to me is that we don't have this kind of table associated file extensions and mimeTypes already somewhere in our code... or in libmagic.
> 
> Good point. There is already `extMimeTypes` in`src/zimwriterfs/tools.cpp`, and `zimwriterfs` uses libmagic as a fallback.
> 
> I did not reuse `extMimeTypes` directly because it selects one preferred MIME type when creating an entry, while validation needs to accept multiple valid MIME types for some extensions. For example, `text/javascript` is the current MIME type for `.js`, but existing archives may legitimately use the deprecated `application/javascript` alias.
> 
> The existing table is also private to the optional `zimwriterfs` target. libmagic identifies MIME types from file contents, but does not provide an extension-to-compatible-MIME-types relation.
> 
> Alternatively, I can move the compatibility information into shared tools while keeping separate APIs for selecting a preferred MIME type and validating compatible MIME types.

I would propose to:
* Extend `extMimeTypes` with the necessary supported file extensions (still use the first one as prefered)... adapt the code in `zimwriterfs`
* Move `extMimeTypes` to `src/tools.cpp`
* We probably need now to find a quality source to complete `extMimeTypes` table (maybe we can take/extract it from libmagic source code)

@veloman-yunkan Any thoughs?

kelson42:
@Nidhi-Gahlawat I'm unsure about the status here, it seems you have do the necessary to have only one extension/mime-type table for all the tools, but I still see only one extension per mime-type?

I have also made a small research to have an up2date database of mimetype/extension and I found `https://github.com/patrickmccallum/mimetype-io/blob/master/src/mimeData.json`. Would that be possible to just transform this to replace or custom made C++ structure?

Nidhi-Gahlawat:
I moved the MIME mapping into the shared `tools.cpp` and grouped related extensions and aliases under each preferred MIME type. I used the linked dataset as a guide, but kept the list focused on common extensions to avoid false warnings.

kelson42:
@Nidhi-Gahlawat Thank you. A problem we still have is about undocumented mime-types/extensions. Beside the mismatches which shoukd be treated as errors, please create a warning in the log if this happens, so we get a change to identify our blind spots.

kelson42:
@veloman-yunkan I think this PR is ready for technical review.

benoit74 (commented):
If code does what PR description says, this LGTM

veloman-yunkan (commented):
I looked rather briefly and the preliminary feedback is this:

1. The PR must start with a refactoring step (making `extMimeTypes` available for using outside `zimwriterfs`).
2. The commit "Expand shared MIME type mappings" needs to be explained (using comments documenting `MimeTypeMapping`) and justified (which problem that design is trying to solve).

src/zimcheck/checks.cpp
```diff
@@ -37,27 +38,36 @@ std::unordered_map<TestType, std::pair<LogTag, std::string>> errormapping = {
...
+    { MsgId::DANGLING_LINKS,   { TestType::URL_INTERNAL, "Dangling link(s) in article '{{&path}}':\n{{#links}}  - '{{&value}}' (resolves to '{{&normalized_link}}')\n{{/links}}" } },
+    { MsgId::EXTERNAL_LINK,    { TestType::URL_EXTERNAL, "{{&link}} is an external dependence in article {{&path}}" } },
+    { MsgId::EMPTY_LINKS,      { TestType::URL_EMPTY, "Found {{&count}} empty links in article: {{&path}}" } },
+    { MsgId::REDUNDANT_ITEMS,  { TestType::REDUNDANT, "{{&path1}} and {{&path2}}" } },
+    { MsgId::METADATA,         { TestType::METADATA, "{{&error}}" } },
+    { MsgId::REDIRECT_LOOP,    { TestType::REDIRECT, "Redirect loop exists from entry {{&entry_path}}\n"  } },
+    { MsgId::MISSING_FAVICON,  { TestType::FAVICON, "Favicon is missing" } },
+    { MsgId::MIME_TYPE_MISMATCH,
+        { TestType::MIME_TYPE,
+            "Entry {{&path}} has MIME type {{&mime_type}}, which is incompatible with the .{{&extension}} extension" } },
```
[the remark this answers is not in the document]
Nidhi-Gahlawat: Fixed in 463e91c: MIME_TYPE_MISMATCH now has an explicit LogTag::WARNING, and text/JSON tests verify that mismatches remain warnings and do not fail the check.

src/zimcheck/checks.cpp
```diff
@@ -289,6 +305,32 @@ void test_mainpage(const zim::Archive& archive, ErrorLogger& reporter) {
...
+    if (extension.empty()) {
+        return;
+    }
+
+    if (!isMimeTypeExtensionKnown(extension) || !isMimeTypeKnown(mimeType)) {
+        reporter.addMsg(MsgId::MIME_TYPE_UNKNOWN,
+                        {{"path", path},
+                         {"mime_type", mimeType},
+                         {"extension", extension}});
+        return;
```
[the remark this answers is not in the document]
Nidhi-Gahlawat: Fixed in 463e91c: unknown extensions now return without logging, while a known extension with an undocumented MIME type still emits a warning. The tests now cover both cases separately.

Nidhi-Gahlawat:
@veloman-yunkan Addressed the latest review in 463e91c and rewrote the branch into two buildable commits:

1. `217851a Share MIME type mappings across tools` is now a behavior-preserving refactor that moves the existing writer mapping into shared tools without changing preferred MIME values.
2. `463e91c Check MIME types against filename extensions` adds the checker and documents why `MimeTypeMapping` separates the canonical writer MIME type from accepted aliases.

src/zimcheck/checks.cpp
```diff
@@ -37,27 +38,37 @@ std::unordered_map<TestType, std::pair<LogTag, std::string>> errormapping = {
...
     { TestType::URL_EMPTY,     {LogTag::WARNING, "Empty link"}},
     { TestType::REDIRECT,      {LogTag::ERROR, "Redirect Loop"}},
+    { TestType::MIME_TYPE,     {LogTag::ERROR, "MIME type"}},
 };
 
 struct MsgInfo
 {
   TestType check;
   std::string msgTemplate;
+    std::optional<LogTag> tag;
```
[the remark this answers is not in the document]
kelson42: @veloman-yunkan I think we agreed that we now support C++17 and `std::optional` seems to come with C++17. Do you confirm we are allowed to use it because it won't brake on kiwix-build CI/CD?

kelson42:
@veloman-yunkan Are we good to merge?

src/zimwriterfs/tools.cpp
```diff
@@ -118,8 +78,7 @@ inline std::string inflateString(const std::string& str)
 
 std::string getMimeTypeFromExtension(const std::string& extension)
 {
-  const auto it = extMimeTypes.find(asciitolower(extension));
-  return it != extMimeTypes.end() ?  it->second : "";
+  return getPreferredMimeTypeForExtension(extension);
 }
```
veloman-yunkan, line 82: Why don't you get rid of this function?

src/tools.h
```diff
@@ -182,6 +182,7 @@ std::string computeAbsolutePath(const std::string& path,
 bool fileExists(const std::string& path);
 bool isDirectory(const std::string &path);
 std::string getFileExtension(std::string_view path);
+std::string getPreferredMimeTypeForExtension(std::string_view extension);
```
veloman-yunkan, line 185: The return type of this function can be `const std::string&`

src/tools.cpp
```diff
@@ -87,51 +93,118 @@ std::string getFileExtension(std::string_view path) {
...
-  {"wasm",       "application/wasm"}
+struct MimeTypeMapping
+{
+  // MIME type emitted by zimwriterfs for all extensions in this mapping.
+  std::string_view mimeType;
+  // Filename extensions conventionally associated with the MIME type.
+  std::vector<std::string_view> extensions;
+  // Valid legacy or alternative MIME types accepted by zimcheck.
+  std::vector<std::string_view> alternatives;
+};
```
veloman-yunkan, line 104: Why separating `alternatives` into a field of their own? Can't we merge them with `mimeType` into a single collection `std::vector<std::string_view> mimeTypes` and treat the first (0th) entry specially (if at all needed)?

That would significantly simplify the data table below.

src/tools.cpp
```diff
@@ -87,51 +93,118 @@ std::string getFileExtension(std::string_view path) {
...
+  const auto normalizedExtension = asciitolower(std::string(extension));
+  for (const auto& mapping : mimeTypeMappings) {
+    if (std::find(mapping.extensions.begin(), mapping.extensions.end(),
+                  normalizedExtension)
+        != mapping.extensions.end()) {
+      return std::string(mapping.mimeType);
+    }
+  }
+  return "";
+}
```
veloman-yunkan, line 166: This can result in an unacceptable slow-down of `getPreferredMimeTypeForExtension()`. Please construct (during static initialization) the old style `extMimeTypes` from `mimeTypeMappings` and leave `getPreferredMimeTypeForExtension()` unchanged.

veloman-yunkan (changes requested):
The second commit is too big a change and has to be decomposed into smaller steps.

src/zimcheck/checks.cpp
veloman-yunkan, line 48: I don't get why this is needed. You map `TestType::MIME_TYPE` in `errormapping` to `LogTag::ERROR` and then override it with `LogTag::WARNING` for all messages associated with `TestType::MIME_TYPE`. Why don't you instead just map `TestType::MIME_TYPE` in `errormapping` to  `LogTag::WARNING`?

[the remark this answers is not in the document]
veloman-yunkan: @kelson42 It looks like [we don't need it at all](https://github.com/openzim/zim-tools/pull/540/changes#r4053252200).

PR #541, opened 2026-08-27
Zimcheck local anchor validation

### LINKED ISSUE #120: zimcheck local anchor validation
Add the functionality to zimcheck to validate local html anchors

See #86

Pranavjeet-Naidu:
Fixes #120

- Validate `#fragment` on internal links as part of the existing `-U` / `--url_internal` check
- Same-page links (#section) and cross-page links (page.html#section) are checked after path resolution
- Report broken anchors with `BROKEN_ANCHOR` (one line per link, like `OUTOFBOUNDS_LINK`)
- Add `extract_link_fragment`, `generic_getAnchorIds` in `tools.cpp` (same lightweight HTML scan as `generic_getLinks`)

NOTE: 
- Query strings (`?foo=bar`) are still ignored for path checks only; this PR does not validate query params
- Empty fragment (`href="#"`) is treated as valid (top of page)
- HTML parsing limitations match `generic_getLinks` (no full parser, comments/code blocks may false-match)

@kelson42 we will need to regen poor.zim on a linux machine, I generated it on my mac and tests failed because of the difference in BSD sed and MacOS sed. 
P.S can we work on changes for [sed usage](https://stackoverflow.com/questions/16745988/sed-command-with-i-option-in-place-editing-works-fine-on-ubuntu-but-not-mac/16746032) in the test scripts, maybe later as part of a separate PR?


kelson42:
@Pranavjeet-Naidu Thank you very much for your PR, we will look at it.
@veloman-yunkan Can you please help with the questions and `poor.zim`?
@benoit74 A small ping as this PR is pretty relevant for you IMHO

benoit74:
Thank you! 

Sorry for not reacting to the issue earlier, "broken anchors" is a weird concept IMHO because AFAIK nothing in the HTML spec mandates the fragment to point to an anchor. Fragments are very commonly interpreted by JS code for instance. 

I'm not even sure it is a very relevant warning, because I'm quite sure it will generate noise in quite a lot of situations. Even if the fragment was meant to lead to an anchor and this anchor is missing, there is no significantly broken user experience. I don't feel like it should be zimcheck responsibility to check this level of detail.

So IMHO "broken anchors" should at most be a warning, not an error. Or at least this check should not activated by default in zimcheck.

kelson42:
> So IMHO "broken anchors" should at most be a warning, not an error. Or at least this check should not activated by default in zimcheck.

Lets do so, seems a careful approach.

Pranavjeet-Naidu:
> I'm not even sure it is a very relevant warning, because I'm quite sure it will generate noise in quite a lot of situations. Even if the fragment was meant to lead to an anchor and this anchor is missing, there is no significantly broken user experience. I don't feel like it should be zimcheck responsibility to check this level of detail.

@benoit74 thanks for bringing this up, that sounds right. 

> So IMHO "broken anchors" should at most be a warning, not an error. Or at least this check should not activated by default in zimcheck.
yes, I've changed it to be a warning instead, thank you. 

@kelson42 sorry for not responding sooner, but please review whenever possible, thank you.

kelson42:
@veloman-yunkan Can you please code review the feature?

src/tools.cpp
```diff
@@ -445,6 +446,96 @@ std::vector<html_link> generic_getLinks(const std::string& page)
...
+            if (hi >= 0 && lo >= 0) {
+                output.push_back(static_cast<char>((hi << 4) | lo));
+                i += 2;
+                continue;
+            }
+        }
+        output.push_back(input[i]);
+    }
+    return output;
+}
```
veloman-yunkan, line 476: Do you really think that `zimcheck` could get along without such a function? Our version is called `decodeUrl()`. It's not perfect but better reuse it and fix any problems with it rather than introducing a functional duplicate.

src/tools.cpp
```diff
@@ -445,6 +446,96 @@ std::vector<html_link> generic_getLinks(const std::string& page)
...
+    int nrsteps = 0;
+    std::string::size_type pos = 0;
+    while((pos = input.find("../", pos)) != std::string::npos) {
+        nrsteps++;
+        pos += 3;
+    }
+
+    return nrsteps >= (nr + std::count(base.cbegin(), base.cend(), '/'));
+}
+
```
veloman-yunkan, line 538: This function was deleted recently and is (likely, unintentionally) resurrected in your PR.

Pranavjeet-Naidu:
@kelson42 @veloman-yunkan: 
 just an fyi, in the recent commit I've committed the regenerated poor.zim just to ensure tests pass, not sure if this is fine.

src/tools.cpp
```diff
@@ -445,6 +446,96 @@ std::vector<html_link> generic_getLinks(const std::string& page)
...
+std::string extract_link_fragment(const std::string& url)
+{
+    const auto pos = url.find('#');
+    if (pos == std::string::npos)
+        return "";
+    return url_percent_decode(url.substr(pos + 1));
+}
+
+std::unordered_set<std::string> generic_getAnchorIds(const std::string& page)
```
veloman-yunkan, line 488: `std::unordered_set` uses too much memory. You may use it internally for speed, but better return the result as a sorted `std::vector` of `std::string`s (better encapsulated into a small helper template class `ImmutableSet<T>` with the main API method `bool contains(const T& ) const`). This will enable us to afford a larger cache for anchor ids.

src/tools.cpp
```diff
@@ -445,6 +446,96 @@ std::vector<html_link> generic_getLinks(const std::string& page)
...
+        while (*p == ' ') p++;
+        if (*p++ != '=') continue;
+        while (*p == ' ') p++;
+        char delimiter = *p++;
+        if (delimiter != '\'' && delimiter != '"') continue;
+        const char* valStart = p;
+        while (*p != delimiter) p++;
+        ids.insert(decodeHtmlEntities(std::string(valStart, p)));
+        p += 1;
+    }
```
veloman-yunkan, line 513: You seem to have copied this from an older version of `generic_getAnchorIds()` which has recently been improved.

src/zimcheck/checks.cpp
```diff
@@ -413,9 +442,17 @@ void ArticleChecker::check_internal_links(zim::Item item, const LinkCollection&
...
-        if (l.link.front() == '#' || l.link.front() == '?') continue;
+        if (l.link.front() == '?') continue;
         if (l.isInternalUrl() == false) continue;
 
+        const auto fragment = extract_link_fragment(l.link);
+
+        if (l.link.front() == '#') {
+            if (!has_anchor(path, fragment)) {
+                reporter.addMsg(MsgId::BROKEN_ANCHOR, {{"link", l.link}, {"path", path}});
+            }
```
veloman-yunkan, line 453: Please respect the [recently added `--quick` mode](https://github.com/openzim/zim-tools/pull/505) of `zimcheck`.

src/zimcheck/checks.cpp
```diff
@@ -413,9 +442,17 @@ void ArticleChecker::check_internal_links(zim::Item item, const LinkCollection&
             nremptylinks++;
             continue;
         }
-        if (l.link.front() == '#' || l.link.front() == '?') continue;
+        if (l.link.front() == '?') continue;
         if (l.isInternalUrl() == false) continue;
 
+        const auto fragment = extract_link_fragment(l.link);
```
veloman-yunkan, line 448: Have you considered enhancing `InternalLinkResolver` so that it returns a triplet `(path, fragment, query)`? I think that under such an approach the code will be somewhat simpler.

src/zimcheck/checks.cpp
```diff
@@ -429,6 +466,10 @@ void ArticleChecker::check_internal_links(zim::Item item, const LinkCollection&
         }
 
         groupedLinks[resolved].push_back(l.link);
+
+        if (!fragment.empty() && is_valid_internal_link(resolved) && !has_anchor(resolved, fragment)) {
+            reporter.addMsg(MsgId::BROKEN_ANCHOR, {{"link", l.link}, {"path", path}});
+        }
```
veloman-yunkan, line 472: 1. If the current article contains multiple occurrences of the same cross-page link with a broken anchor, each of them will be reported separately. Ideally this check must be moved to the other overload of `ArticleChecker::check_internal_links()` but this means that `html_link` must be enhanced (a `fragment` data member must be added to it).

2. Even after above optimization the performance impact of the new check may be significant on huge ZIM files with the count of pages significantly exceeding the size of the new cache for anchors. I think that checking of anchors must be optional.  

2. `--quick` mode must be respected.

veloman-yunkan (changes requested):
I think that the new check must be optional. It must also respect the recently added [`--quick` mode](https://github.com/openzim/zim-tools/pull/505) of `zimcheck`.

veloman-yunkan:
> @veloman-yunkan Can you please help with the questions and `poor.zim`?

A longer-term solution to the problem with the maintenance of the test data has to be found. Need to think about it.

kelson42:
> I think that the new check must be optional. It must also respect the recently added [`--quick` mode](https://github.com/openzim/zim-tools/pull/505) of `zimcheck`.

Why it should be optional? I really disagree with that.

veloman-yunkan:
@kelson42 The new check [will come with a noticeable impact on performance](https://github.com/openzim/zim-tools/pull/541#discussion_r3958310533) (for large ZIM files). Also it [may generate a lot of noise](https://github.com/openzim/zim-tools/pull/541#issuecomment-5449945782). I think that those two reasons provide enough justification for introducing it as an optional check (off by default, which can be changed after our fears prove unsubstantiated).

kelson42:
@Pranavjeet-Naidu Have you made some benchmark with/without on a small Wikipedia file to verify it performs well? Maybe you alreqdy have such a benchmark?

src/zimcheck/checks.cpp
Pranavjeet-Naidu: I've added `ResolvedLink { path, query, fragment }` and `InternalLinkResolver::resolve()`. 
`resolveLinkTarget()` is now `return resolve(url).path` so the old tests stay.

src/zimcheck/checks.cpp
Pranavjeet-Naidu: `--quick` now stops after the first `BROKEN_ANCHOR` on that article (reportedAnchor). follows the same idea as dangling links.

Pranavjeet-Naidu:
> @Pranavjeet-Naidu Have you made some benchmark with/without on a small Wikipedia file to verify it performs well? Maybe you alreqdy have such a benchmark?

```bash
$ZIM=.worktrees/wikipedia_en_100_2026-08.zim
time build/src/zimcheck/zimcheck -U --threads=1 "$ZIM" >/dev/null
time build/src/zimcheck/zimcheck --anchors --threads=1 "$ZIM" >/tmp/anchors.out
grep -c "Broken Anchor" /tmp/anchors.out

real    0m0.323s
user    0m0.172s
sys     0m0.066s

real    0m0.584s
user    0m0.516s
sys     0m0.064s
2706
bash-5.3$ 


```
given this test, we should go with it being optional and being off by default.

kelson42:
@veloman-yunkan I'm not sure regarding this optional/default topic, can you please complete the review while I think about it?

Issue #549, opened 2026-09-01
Functionality of the `--force` option of `zimsplit` must be reinstated


kelson42:
`zimsplit` has a document option `--force` however it always works as if having been invoked with that option.

Follow-up of https://github.com/openzim/zim-tools/pull/537#discussion_r3831469793

kelson42:
@Nidhi-Gahlawat Any chance you could quickly make a PR for this issue?

veloman-yunkan:
@kelson42 The problem with `zimsplit --force` reported in #537 was a different one. See https://github.com/openzim/zim-tools/pull/537#discussion_r3828469502.

Nidhi-Gahlawat:
@kelson42, @veloman-yunkan PR #551  makes an infeasible part size a hard error regardless of `--force`. Consequently, `--force` and `ZimSplitter::check()` become redundant. Is this issue asking us to remove the option and its documentation entirely?

kelson42:
@Nidhi-Gahlawat Thank you for your comment, but please lets do that on the PR itself! #551 is not even a fix for this PR!

But would you like to implement this one (yes I know it is trivial)?

veloman-yunkan:
The title and description of this PR don't make much sense and must be updated.

PR #551, opened 2026-09-01
Fix #550: enforce minimal chunk size in zimsplit

### LINKED ISSUE #550: Properly check minimal chunk size in `zimplit`
The current check will happily accept 1 byte parts. This is not feasible and `zimsplit` should complain accordingly.

This is a follow-up of https://github.com/openzim/zim-tools/pull/537#discussion_r3839203702

LJX1021403:
Added a check in the `ZimSplitter` constructor to compute the minimum required chunk size (the largest chunk, which is either metadata + first cluster or a single cluster). If the requested `--size` is smaller than this minimum, the program throws an exception and exits, regardless of `--force`. This prevents creating invalid ZIM parts.

### Changes
- Compute all chunk boundaries using `getOffsets()`.
- Determine the maximum chunk size as the minimum possible part size.
- Throw `std::invalid_argument` if `maxPartSize` is smaller than this minimum.
- The check happens before any file writing, so it cannot be bypassed by `--force`.

### Testing
- Verified that `--size 1` on `test/data/zimfiles/good.zim` now reports:
  `part size must be at least 43922 bytes (the size of the largest chunk)`
  and exits without creating parts.
- Existing test suite (`meson test`) passes.

Fixes #550

### DIFF
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -65,16 +65,14 @@ class ZimSplitter
         // Compute the largest chunk size, where a chunk is either the
         // metadata + first cluster or a single cluster
         auto offsets = getOffsets();
-        if (offsets.empty()) {
+        if (offsets.size()<2) {
             throw std::runtime_error("ZIM file contains no clusters to split");
         }
 
-        minRequiredChunkSize = 0;
-        zim::offset_type last = 0;
-        for (auto offset : offsets) {
-            auto chunkSize = static_cast<zim::size_type>(offset - last);
-                minRequiredChunkSize = std::max(minRequiredChunkSize, chunkSize);
-                last = offset;
+         minRequiredChunkSize = static_cast<zim::size_type>(offsets[1]);
+         for (std::size_t i = 2; i < offsets.size(); ++i) {
+             auto chunkSize = static_cast<zim::size_type>(offsets[i] - offsets[i - 1]);
+             minRequiredChunkSize = std::max(minRequiredChunkSize, chunkSize);
         }
         batch_buffer = new char[BUFFER_SIZE];
     }
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -20,6 +20,7 @@
 #include <fstream>
 #include <sstream>
 #include <stdexcept>
+#include <algorithm>
 
 #define ZIM_PRIVATE
 #include <zim/archive.h>
@@ -59,6 +60,27 @@ class ZimSplitter
         currentPartSize(0)
       {
         validatePartSize(maxPartSize, archive.getFilesize());
+        // Compute the largest chunk size, where a chunk is either the
+        // metadata + first cluster or a single cluster
+        auto offsets = getOffsets();
+        if (offsets.empty()) {
+            throw std::runtime_error("ZIM file contains no clusters to split");
+        }
+
+        zim::size_type minRequiredSize = 0;
+        zim::offset_type last = 0;
+        for (auto offset : offsets) {
+            auto chunkSize = static_cast<zim::size_type>(offset - last);
+                minRequiredSize = std::max(minRequiredSize, chunkSize);
+                last = offset;
+        }
+
+        if (maxPartSize < minRequiredSize) {
+              throw std::invalid_argument(
+                "part size must be at least " + std::to_string(minRequiredSize) +
+                " bytes (the size of the largest cluster)");
+        }
+
         batch_buffer = new char[BUFFER_SIZE];
     }
 
--- a/src/zimsplit.cpp
+++ b/src/zimsplit.cpp
@@ -21,6 +21,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <algorithm>
+#include <iostream>
 
 #define ZIM_PRIVATE
 #include <zim/archive.h>
@@ -47,6 +48,7 @@ class ZimSplitter
     std::ofstream ofile;
     std::string part_name;
     zim::size_type currentPartSize;
+    zim::size_type minRequiredChunkSize;
     char* batch_buffer;
 
   public:
@@ -67,22 +69,17 @@ class ZimSplitter
             throw std::runtime_error("ZIM file contains no clusters to split");
         }
 
-        zim::size_type minRequiredSize = 0;
+        minRequiredChunkSize = 0;
         zim::offset_type last = 0;
         for (auto offset : offsets) {
             auto chunkSize = static_cast<zim::size_type>(offset - last);
-                minRequiredSize = std::max(minRequiredSize, chunkSize);
+                minRequiredChunkSize = std::max(minRequiredChunkSize, chunkSize);
                 last = offset;
         }
-
-        if (maxPartSize < minRequiredSize) {
-              throw std::invalid_argument(
-                "part size must be at least " + std::to_string(minRequiredSize) +
-                " bytes (the size of the largest cluster)");
-        }
-
         batch_buffer = new char[BUFFER_SIZE];
     }
+
+    zim::size_type getMinRequiredChunkSize() const { return minRequiredChunkSize; }
 
     ~ZimSplitter() {
         close_file();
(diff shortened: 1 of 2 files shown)


LJX1021403:
Thanks for the feedback. I see the discussion in #549 about the --force option. I'm happy to adjust the implementation to align with whatever behavior the maintainers decide is best. Please let me know if you'd like me to make the minimal chunk size check respect --force (e.g., only enforce it when --force is not passed), or if another approach is preferred.

kelson42:
The `--force` command line argument can be used to `Create zim parts even if it is impossible to have all part size smaller than requested`. As a consequence:
* The `--force` option should not allow to create broken ZIM file
* If the `--force` option is not set and it is impossible to respect the minimum chunk size, then `zimsplit` should fail with appropriate error
* If the size is so small than all chunks will be bigger than the chunk size given, then just stop with error. Otherwise go to the last option.
* If the `--force` is set, then we print a nice warning that we are not able to be so small but will use this other chunk size and we move forward.

Hope this helps.

@Nidhi-Gahlawat Glad to have you reviewing this PR :)

LJX1021403:
Thanks for the detailed feedback. I've updated the implementation:
- Without `--force`, the program now reports an error and exits when the requested size is smaller than the minimum feasible chunk size.
- With `--force`, it prints a warning and continues, splitting along cluster boundaries to avoid corrupt ZIM files.

I've also tested both cases manually on `test/data/zimfiles/good.zim`. Please let me know if any further changes are needed.

kelson42:
@LJX1021403 Thank you, seems to work fine to me.
@veloman-yunkan @Nidhi-Gahlawat Glad to get your review

kelson42 (approved):
LGTM

LJX1021403:
Thanks for the detailed review! I've updated the calculation as suggested:
- The first chunk is now treated as `offsets[1]` (metadata + first cluster).
- Subsequent chunks are computed as `offsets[i] - offsets[i-1]`.
- The empty check now uses `offsets.size() < 2`.

I've re-tested both `--force` and non-`--force` cases manually. Please let me know if anything else is needed.

Nidhi-Gahlawat:
LGTM. A small regression test in zimsplit-test.cpp could be added later for the --force and too-small-size paths.

LJX1021403:
Thanks for the suggestion! I've added two regression tests in `zimsplit-test.cpp`:
- one verifies the program returns an error when `--size` is too small and `--force` is not used;
- another verifies it continues (with a warning) when `--force` is used.

The test suite passes. Please let me know if any further changes are needed.

test/zimsplit-test.cpp
```diff
@@ -109,3 +130,26 @@ TEST(ZimSplitSize, ValidatesPartSizeAgainstArchive)
...
+{
+    const char* argv[] = {
+        "../build/src/zimsplit",
+        "--size", "1",
+        "--force",
+        "data/zimfiles/good.zim",
+        nullptr
+    };
+    EXPECT_EQ(run_command(argv), 0);
+}
```
veloman-yunkan, line 155: This relies on the build method used. It won't work in our CI.

Besides, this is fragile, since tests can be built and run without rebuilding the executables (which can result in tests locally passing with an actual bug having been introduced in the code but not yet having made its way into the binary).

test/zimsplit-test.cpp
```diff
@@ -23,6 +25,25 @@ constexpr auto TOO_LARGE_PART_SIZE_ERROR
...
+        _exit(127); // exec 失败
+    }
+    // 父进程：等待子进程结束
+    int status = 0;
+    waitpid(pid, &status, 0);
+    if (WIFEXITED(status)) {
+        return WEXITSTATUS(status);
+    }
+    return -1;
+}
```
veloman-yunkan, line 45: Avoid this. See how it's done for `zimcheck`.

veloman-yunkan (commented):
First of all we must restore the documented behaviour of the `--force` option. Then we can set some practical lower limit (e.g. 100MB) on the value of the `--size` option as a fool-proofness measure. `zimsplit` is intended to split large ZIM files into manageable pieces, rather than for fooling around. That limit can be overridden via `--force` similar to its effect in other cases.

kelson42:
@veloman-yunkan Do we have any external blocker?
@LJX1021403 Any feedback on the comment review?

veloman-yunkan:
> @veloman-yunkan Do we have any external blocker?

@kelson42 No, #552 has been merged, so the minimal chunk-size can now be implemented as a stupid hard-coded value as proposed in https://github.com/openzim/zim-tools/pull/551#pullrequestreview-5168716284. Just need to decide what that value should be.

LJX1021403:
Sorry for the delayed reply. I've merged upstream/main and resolved the conflicts. The PR now:

1. Keeps the upstream `--force` semantics (`check()` still runs when `--force` is not used).
2. Adds a 100MB practical lower limit (`MIN_PRACTICAL_PART_SIZE`) on `--size`:
   - Without `--force`: sizes below 100MB are rejected with an error.
   - With `--force`: a warning is printed and splitting continues.
3. Refactors `main` into a testable `zimsplit()` function.
4. Rewrites tests to call `zimsplit()` directly with captured stdout/stderr (following the `zimcheck` pattern), removing the `fork/exec` approach.
5. Adds `-DZIMSPLIT_TEST` to exclude `main` when building the test.

The test suite passes, and I've manually verified both `--force` and non-`--force` cases. Please let me know if the 100MB value should be adjusted or if any further changes are needed.

kelson42:
@LJX1021403 Please rebase on main, not merge.

LJX1021403:
Done — rebased onto upstream/main. The branch now contains a single clean commit, no merge commits.

kelson42:
@LJX1021403 thx. I will request a new review of @veloman, but the fax that you have squash everything in one commit makes the new review pass complicated, because we don't see what you havenfixed. Next time just add new commits with your fixes (one commit per fix).

LJX1021403:
Thanks for the feedback. I've replaced the hard-coded 100MB limit with the actual largest chunk size (computed from offsets), matching the run()/check() logic. Tests now assert the exact error messages and the warning output. Changes are in a new commit.

veloman-yunkan:
@LJX1021403 Copilot has stupidly reviewed your PR based on the old description of the PR and having ignored the discussion that followed. Please undo the last change.

test/zimsplit-test.cpp
```diff
@@ -109,3 +144,19 @@ TEST(ZimSplitSize, ValidatesPartSizeAgainstArchive)
...
+    EXPECT_NE(ret, 0);
+}
+
+TEST(ZimSplit, WarnsAndContinuesWithForce)
+{
+    CapturedStdout out;
+    CapturedStderr err;
+    int ret = zimsplit({"zimsplit", "--size", "1", "--force", "data/zimfiles/good.zim"});
+    EXPECT_EQ(ret, 0);
+}
```
veloman-yunkan, line 134: Why did you introduce (copy from `zimcheck`) `CapturedStdout` and `CapturedStderr` if you are not going to check the content of the captured streams?

test/zimsplit-test.cpp
```diff
@@ -21,6 +26,36 @@ namespace
...
+struct CapturedStdout : CapturedStdStream
+{
+  CapturedStdout() : CapturedStdStream(std::cout) {}
+};
+
+struct CapturedStderr : CapturedStdStream
+{
+  CapturedStderr() : CapturedStdStream(std::cerr) {}
+};
+
```
veloman-yunkan: Don't duplicate code, share it instead.

src/zimsplit.cpp
```diff
@@ -225,3 +245,10 @@ int main(int argc, char* argv[])
   return 0;
 }
 
+#ifndef ZIMSPLIT_TEST
+int main(int argc, char* argv[])
+{
+  std::vector<const char*> args(argv, argv + argc);
+  return zimsplit(args);
+}
+#endif
```
veloman-yunkan: Create a new file `src/zimsplit_main.cpp` instead of this.

veloman-yunkan (changes requested):
Please structure the PR as the following commits:

1. Moving `CapturedStdXXX` testing utilities into a shared file `test/testing_helpers.{h,cpp}`
2. Exraction of the `zimsplit` function
2. Add a few test cases of `zimsplit` checking the current implementation of handling of various command lines. Please cover various scenarios including those not directly related to the ticket being addressed. 
4. Fix the ticket and update the related tests introduced in the previous step.

LJX1021403:
I've restructured the PR into the requested commits:

1. Extract testing helpers into a shared file (`test/testing_helpers.{h,cpp}`).
2. Extract `zimsplit` function into a separate translation unit (`src/zimsplit_main.cpp` for `main`).
3. Add command-line handling tests covering various scenarios.
4. Assert specific error and warning messages in `zimsplit` tests.

I've also removed the `#ifndef ZIMSPLIT_TEST` guard since `main` is now in a separate file.

The 100MB lower limit fix is included. Tests pass locally. Please let me know if anything else needs adjusting.

test/testing_helpers.cpp
```diff
@@ -0,0 +1 @@
+// Shared testing helpers are defined inline in testing_helpers.h
```
veloman-yunkan, line 1: We don't need this file as long as it is going to remain empty.

veloman-yunkan (changes requested):
You added the 4 commits on top of the original commit. The PR should be structured as requested in the previous review.

PR #552, opened 2026-09-02
Restore zimsplit --force validation behavior

### LINKED ISSUE #549: Functionality of the `--force` option of `zimsplit` must be reinstated

Nidhi-Gahlawat:
## Summary

- Read `--force` as a boolean so validation runs when the option is absent.
- Skip archive-size and split-feasibility checks only when `--force` is supplied.
- Make `ZimSplitter::check()` return `true` when the requested split is valid.
- Document that forced operation may exceed the requested size or produce a single part.

The regression was caused by testing the docopt value object directly instead of calling `asBool()`, which caused zimsplit to bypass its feasibility check even when `--force` was not supplied.

## Testing

- `meson test -C build --print-errorlogs`
- Manually verified an impossible one-byte split fails without `--force` and succeeds with it.
- Manually verified an archive-sized part fails without `--force`, succeeds with it, and produces output byte-for-byte identical to the input.

Fixes #549


kelson42 (approved):
LGTM

@veloman-yunkan Can you please do the technical review?

src/zimsplit.cpp
```diff
@@ -58,7 +61,7 @@ class ZimSplitter
         ifile(fname, std::ios::binary),
         currentPartSize(0)
       {
-        validatePartSize(maxPartSize, archive.getFilesize());
+        validatePartSize(maxPartSize, archive.getFilesize(), force);
```
veloman-yunkan, line 64: I'd rather perform validation only when `force == false`.

src/zimsplit.cpp
```diff
@@ -197,11 +202,12 @@ int main(int argc, char* argv[])
...
     if (args["--size"])
         size = parseByteSize(args["--size"].asString());
+    const bool force = args["--force"].asBool();
 
     // initalize app
-    ZimSplitter app(args["<file>"].asString(), prefix, size);
+    ZimSplitter app(args["<file>"].asString(), prefix, size, force);
 
-    if (!args["--force"] && app.check()) {
+    if (!force && app.check()) {
```
veloman-yunkan, line 210: I don't understand why we had a bug with `--force`.

BTW, the semantics of `app.check()` is counterintuitive - it returns `true` in case of an error. I think we should fix that.

veloman-yunkan (changes requested):
Commit description must be updated to properly reflect the fix (see updated description of the ticket).

src/zimsplit.cpp
Nidhi-Gahlawat: Updated in ad67e11: archive-size validation is now called only when force is false; the generic validatePartSize helper no longer accepts a force flag.

src/zimsplit.cpp
Nidhi-Gahlawat: Updated in ad67e11: the bug came from testing the docopt value object directly instead of args["--force"].asBool(), which caused checks to be skipped when the flag was absent. ZimSplitter::check() now returns true for a valid split, and the caller uses !app.check() for the error path.

Nidhi-Gahlawat:
@veloman-yunkan The requested changes are addressed in ad67e11: validation now runs only when `force == false`, `ZimSplitter::check()` returns true for a valid split, and the commit/PR descriptions explain the actual `--force` regression. Both review threads are resolved. Could you please take another look?

veloman-yunkan (approved):

